PR #20, opened 2015-07-30
Use --relaxed flag when adding URL for sources pulling from master.

ehelms:


setup_sources.sh
```diff
@@ -9,5 +9,10 @@ for spec in $dir/*.spec; do
   sourcebase=$(basename "$source0")
   [ -h $d/$sourcebase ] || continue
   git annex whereis "$d/$sourcebase" 2>/dev/null | grep -q " web" && continue
-  git annex addurl --file "$d/$sourcebase" "$source0"
+  
+  if [[ $2 == '--relaxed' ]]; then
+    git annex addurl --relaxed --file "$d/$sourcebase" "$source0"
+  else
+    git annex addurl $2 --file "$d/$sourcebase" "$source0"
+  fi
```
jlsherrill, line 17: i'm not sure i get the else clause with the $2.  I'd think either the if/else isn't needed with $2, or the else wouldn't contain $2?

jlsherrill:
ACK

PR #23, opened 2015-07-30
fixes #11261 - wait a little after starting httpd

stbenjam:
see https://github.com/Katello/katello/pull/5389

### DIFF
--- a/katello/service-wait
+++ b/katello/service-wait
@@ -12,7 +12,7 @@
 # This simple wrapper works with any service and just pass over the request
 # to /sbin/service, but for few problematic ones it waits until service is
 # fully started or stopped.
-# 
+#
 # To test the implementation use:
 #
 #   service-wait xxx test-wait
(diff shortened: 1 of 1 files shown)


mccun934:
seems entirely reasonable to me, surprised we didn't need this earlier

jlsherrill:
would a wait_for_url work in this case?  Not opposed to the change, just curious

mccun934:
@jlsherrill what URL would we pick?

jlsherrill:
http://localhost:80  ?

I guess it depends on why its failing if you need to try to hit pulp or not

mccun934:
localhost:80 makes sense, that is pretty service agnostic. @stbenjam thoughts?

stbenjam:
@jlsherrill  Yea, I was wondering if we should hit a url or not.  http://localhost:80 hits foreman, and might take a while for the app to load.  Could be many minutes in some cases.

I figured a short sleep was enough, just because the problem here is such a tiny, tiny window,  where the init script creates the lockfile but the pidfile isn't there yet.

jlsherrill:
@stbenjam ahhh if that is the problem then i'm good with this solution,  ACK

stbenjam:
Thanks. I did file a BZ for it too, in case they want to fix the init script.  https://bugzilla.redhat.com/show_bug.cgi?id=1248724

PR #35, opened 2015-08-12
Fixes #11083: Prevent katello-agent from being installed with pulp-consumer-client

ehelms:


stbenjam:
Can you reference #11083 so it gets picked up in redmine?

ehelms:
Good call - updated

stbenjam:
ACK

PR #36, opened 2015-08-12
Update installer

ehelms:


ehelms:
This update is to allow proper use of the Jenkins builder for nightly sources. This both updates the custom source fetcher for Jenkins and adds a tito releaser that makes use of that source fetcher. If a Jenkins job produces an archive that includes a package tarball, that can be used to then build the package. For example:

```
tito release -o /home/ehelms/workspace/katello-packaging/rel-eng/build/ --scratch --arg jenkins_job=release_build_katello_installer --arg jenkins_job_id=lastSuccessfulBuild koji-katello-nightly
```

The release_build_katello_installer Jenkins job runs the generate_source rake command and then archives a tarball of the installer. The above command would grab the last built version and use that as the source.

jlsherrill:
ACK

PR #48, opened 2015-08-14
fixes #11353 - make wait_for_url more reliable

stbenjam:
There's a brief window where tomcat's listening on 8443, but not responding 200 to /candlepin/status.  Sometimes, we end up calling wget in that window, which exits immediately with a failure.

wget does not obey the `--tries` there in this case.

It's generally reproducible if you do this as one command:

``` bash
service tomcat6 stop; service tomcat6 start; /usr/bin/wget --timeout=1 --tries=30 --retry-connrefused -qO- --no-check-certificate https://localhost:8443/candlepin/status; echo $?
```

### DIFF
--- a/katello/service-wait
+++ b/katello/service-wait
@@ -24,6 +24,7 @@ COMMAND=$2
 
 # maximum time to wait (in seconds)
 WAIT_MAX=${WAIT_MAX:-30}
+RETRY_INTERVAL=${RETRY_INTERVAL:-1}
 TOMCAT_PORT=${TOMCAT_PORT:-8443}
 TOMCAT_SERV_PORT=${TOMCAT_SERV_PORT:-8005}
 TOMCAT_TEST_URL=${TOMCAT_TEST_URL:-https://localhost:$TOMCAT_PORT/candlepin/status}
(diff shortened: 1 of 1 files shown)


ehelms:
Looking ahead, do you know if from talking to the Candlepin team there is a potentially more reliable way to handle knowing when the Candlepin "service" is started in the future?

katello/service-wait
```diff
@@ -24,6 +24,7 @@ COMMAND=$2
 
 # maximum time to wait (in seconds)
 WAIT_MAX=${WAIT_MAX:-30}
+RETRY_INTERVAL=${RETRY_INTERVAL:-1}
```
ehelms, line 27: Edit: Ignore that, silly bash syntax always throws me off

ehelms:
The logic makes sense to me, ACK

stbenjam:
I think the URL is fine, it was our use of wget and the subtleties of `--tries`.

wget actually doesn't obey `--tries` for connection refused, despite the comment to the contrary in service-wait. Or 404.  One or the other happens here.

I figured it's better just to do our own loop and forget the subtleties of `wget --tries`.

stbenjam:
Thanks!

PR #50, opened 2015-08-20
[DO NOT MERGE] Adding support for tfm

ehelms:
Note that this currently includes some test scripts and configuration that will be removed before the final merge. However, I figured I'd include them since I am still doing some testing and in case anyone was interested in trying it themselves. The `test_scl.rb` script is designed to configure your environment and build all the necessary packages into a local repository from this repository and link to @domcleal EL7 TFM test repository. So, if you try to test this, I would recommend a clean centos7 VM from katello-deploy, check out this PR and then run `./test_scl.rb`. 

I have run the following tests myself:
- [x] Package build tests with mock
- [x] Package installation on a clean machine `yum install katello`
- [x] Full katello-installer installation + content bats tests
- [x] Package upgrade against 2.3
- [x] Installer upgrade run against 2.3

This will also require:
1. Updating puppet-katello https://github.com/Katello/puppet-katello/pull/76


ekohl:
@ehelms if you use `- [ ]` instead of `[ ]` you get nice checkboxes.

ehelms:
Thanks @ekohl less intuitive, but makes more sense!

ehelms:
I have updated this to handle the removal of haml as a dependency, and some changes to the katello.spec I encountered with respect to upgrades. This should be generally review-able and testable given the test repo from @domcleal using the mock config and my script. Real builds will be able to commence within a few days.

ehelms:
I am going to be taking this PR and splitting it up into individual gem PRs building from the ground up to catch any issues and make the updates a bit cleaner so please be on the look out for those.

ehelms:
Closing as all functionality here has been merged in through various individual PRs

PR #52, opened 2015-08-25
fixes #11568 - adds help scripts to recovery methods

cfouant:


katello/katello-backup
```diff
@@ -1,94 +1,113 @@
 #!/bin/bash
 
 usage(){
-  echo "Usage: katello-backup <backup directory>"
-  echo " eg: $ katello-backup /tmp/katello-backup/"
+  echo "Usage: katello-backup --directory <backup directory>"
+  echo " eg: $ katello-backup --directory /tmp/katello-backup/"
```
komidore64: why are you changing this from being a positional argument? specifying the directory is required, correct?

cfouant: I felt like it was nicer in case down the line we wanted to add other arguments to the script. Open for discussion however.

komidore64: Other arguments are fine. Then wouldn't they just follow before or after the directory argument?

cfouant: They would follow after the directory argument. It's doable, this is easier for me and opens it up for user, but I can change it do katello-backup <directory> --option1 --option2 if you feel that's nicer.

komidore64: What does "opens it up for the user", mean?

komidore64: The point that I am attempting (poorly) to get across is that arguments are for required information that a CLI tool needs, and options are, well, optional.

I realize that Hammer uses options for both required and non-required information completely throwing out the entire notion of positional arguments which drives me bonkers, but that ship has already sailed and I doubt anyone would hear me out to change that now.

If you're wanting to use `--directory DIR` instead of requiring a positional `/path/to/dir` argument for the sake of using the same style as Hammer, then I can live with that.

cfouant: Freedom to add arguments in whatever fashion they would like... 
katello-backup --option1 --directory --option2

cfouant: I'm happy to go to a positional argument, I had asked for feedback on argument vs option and didn't really hear anything back so just went with options.

komidore64: Okay, well if you asked and I didn't chime in. Options are fine.

cfouant: Too late I'm already updating it :)

katello/katello-backup
```diff
@@ -1,94 +1,113 @@
...
+  echo ""
 
-echo ""
-echo ""
-echo "** BACKUP Complete, contents can be found in: $BDIR"
-echo ""
-echo ""
+  # Start services
+  service postgresql restart
+  katello-service restart
```
komidore64: Since we know that these services are being stopped above on lines 41 and 42 maybe we should call `start` on these instead of `restart`?

katello/katello-restore
```diff
@@ -1,77 +1,89 @@
...
-# Restart services
-echo "Restarting all Red Hat Satellite processes"
-service postgresql restart
-katello-service restart
-echo "Done."
-echo ""
+  # Restart services
+  echo "Restarting all Red Hat Satellite processes"
+  service postgresql restart
+  katello-service restart
```
komidore64: `start` vs `restart`?

katello/katello-restore
```diff
@@ -1,77 +1,89 @@
...
+  echo ""
 
-# Restart services
-echo "Restarting all Red Hat Satellite processes"
-service postgresql restart
-katello-service restart
-echo "Done."
-echo ""
+  # Restart services
+  echo "Restarting all Red Hat Satellite processes"
```
stbenjam: Katello\* (or remove product names)

komidore64: :+1:

cfouant:
@stbenjam @komidore64 changes made: altered script to ruby, removed flag for directory (making it an argument in the command rather than a flag), removed references to product names. Let me know what you guys think!

stbenjam:
My feedback looks addressed :smile:  I'll leave it to @komidore64 to comment on the rest

katello/katello-backup
```diff
@@ -1,94 +1,88 @@
...
-# Backup dir
-umask 0027
-BDIR=$1
-mkdir -p $BDIR
-chgrp postgres $BDIR
-chmod 770 $BDIR
-cd $BDIR
-echo "Creating backup folder $BDIR"
-echo ""
+  opts.on_tail("-h", "--help", "Show this help message and exit") do
```
komidore64: `OptionParser` handles help output for you, you shouldn't need to specify it here.

komidore64:
Were you going to parse flags and options at first and then decided not to? It seems like you're using OptionParser only for help output, which isn't wrong. It formats help output quite nicely with little effort from us.

katello/katello-backup
komidore64: i was thinking of clamp, not optionparser. my mistake.

cfouant:
I will be adding options shortly, but wanted to address specific but regarding --help flags first.

katello/katello-backup
cfouant: From what I have seen, you need to define all options, including help.

cfouant: I stand corrected!

katello/katello-backup
```diff
@@ -1,94 +1,83 @@
...
-service postgresql stop
+DIR = ARGV.pop
+if DIR.nil?
+  puts "**** ERROR: Please specify an export directory ****"
+  puts optparse
+  exit(-1)
+else
+  FileUtils.mkdir_p(DIR)
+  puts "Creating backup folder #{DIR}"
+  `chgrp postgres #{DIR}`
```
komidore64: since we're already using `FileUtils` for other things, any objection to changing this to `FileUtils.chown(nil, 'postgres', DIR)`?

komidore64:
looks good to me

ack

komidore64:
@ehelms does this need any magic for SCL shenanigans?

ehelms:
Is it intended to run under the SCL? i.e. can these scripts run under 1.8.7?

komidore64:
all the rest of our ruby is under the scl, correct? would/should this be an exception?

ehelms:
Not all -- katello-service script for example isn't.

cfouant:
hmmm yeah i hadn't thought of that, do i need to change the katello.spec file?

komidore64:
@cfouant i think it's fine like it is.

ehelms:
If it runs on EL6, as is, then we are all good.

komidore64:
@cfouant you can confirm?

cfouant:
let me do another run through since changes made

cfouant:
I had to change one line in the restore scripts from Dir.exist? to File.directory? as the exist method does not exist for the version of ruby in EL6.

EDIT: @ehelms @komidore64 @stbenjam - not sure if I need final ACKs from you guys, given the code change (listed here)

cfouant:
@ehelms @komidore64 @stbenjam do i need final ACKs from you? I did have to make a change to get everything working on EL6

stbenjam:
My feedback was addressed so ACK from me

PR #74, opened 2015-09-01
Stop building katello-sam RPM

ehelms:
This removes katello-sam as an RPM and thus removes dependencies on
rubygem-foreman_sam as well. If this ever needs to be ressurected,
this commit can be used as a template.


ehelms:
This begs the question do we remove the cli and the installer parts as well? @thomasmckay @komidore64 @mccun934

jlsherrill:
ACK  remove all the sam

ehelms:
@jlsherrill updated this to just remove all SAM

jlsherrill:
ACK

PR #76, opened 2015-09-01
fixes #11648 - katello-service: update to start/stop postgresql

bbuckingham:


katello/katello-service
```diff
@@ -14,6 +14,7 @@ SERVICES = {
   'pulp_workers'          => 20,
   'pulp_celerybeat'       => 20,
   'pulp_resource_manager' => 20,
+  'postgresql'            => 20,
```
stbenjam: I don't _think_ there's anything before this point that needs postgres today, but I think it might be better if we do the databases early to avoid such a problem later.  What do you think about including it at the same level as mongod?

bbuckingham: @stbenjam, That seems reasonable.  I can move it to earlier in the flow.

bbuckingham: @stbenjam, Updated

stbenjam:
APJ

ehelms:
Will this break upgrades since we stop services and then attempt to run migrations?

stbenjam:
Hmm, yea, but we should ensure the status of all the databases anyway, PR: https://github.com/Katello/katello-installer/pull/257

If its going to get cherry-picked somewhere we need to make sure it goes together.

bbuckingham:
@stbenjam, Thanks for making the corresponding change to katello-installer!  We can reference both PRs in the BZ to ensure that if the BZ gets pulled in both changes come in.  That said, I do not see it coming in prior to the next 'release'.

stbenjam:
@ehelms mentioned on the other PR that maybe we should include an `--only` option on katello-service, instead of the multiple commands we do in the hooks in that other PR.

Do you think you could add it like we have `--exclude` here? If not I can add it when I have time.

The idea would be we could run:
  `katello-service restart --only postgresql,mongod`

bbuckingham:
@stbenjam, Sure, I can look in to adding an --only option.  I assume we can do that separate and independent of this PR and https://github.com/Katello/katello-installer/pull/257.  Correct?

stbenjam:
It can go here, because we want to use it instead of doing them one-by-one here https://github.com/Katello/katello-installer/pull/257/files#r38475085

bbuckingham:
@stbenjam, Sounds good.

bbuckingham:
@stbenjam, @ehelms, added a --only option to katello-service.  I included it as a separate commit, but can squash, if desired.

ehelms:
Thanks for collaborating on this! ACK for me

stbenjam:
ACK for me too, thanks for catching that @ehelms

bbuckingham:
Thanks for the feedback and review guys!  Merging :)

PR #83, opened 2015-09-25
updating hammer_cli_katello to match gem deps

jlsherrill:


komidore64:
kept you waiting, huh?

ack

PR #85, opened 2015-10-13
Refs #10621: Don't deploy katello.yml during RPM build

ehelms:
The move to the use of SETTINGS included a set of default settings
within the engine declaration of the Katello engine. Thus, we no
longer need to deploy a katello.yaml file as we don't need custom
settings during build.


jlsherrill:
ACK

PR #86, opened 2015-10-15
Fixes #12181 - Restart katello services in right order

adamruzicka:


stbenjam:
Looks good, thanks @adamruzicka

PR #89, opened 2015-10-22
updating nightly to 2.5

jlsherrill:


ehelms:
What about these as well:

katello
katello-agent
katello-repos

jlsherrill:
thx, updated.  

since we don't do builds of nightly builds katello-agent I think it makes more sense to update katello-agent to 2.5 whenever 2.5 is being released (or whenever a new change comes in for katello-agent).

ehelms:
ACK

PR #96, opened 2015-10-27
updating installer for 2.4 rc1

jlsherrill:


katello-installer/katello-installer-base.spec
```diff
@@ -2,7 +2,7 @@
 %undefine scl_prefix
 %global scl_ruby /usr/bin/ruby
 
-%global prever .rc1
+%global prever .rc1.1
```
ehelms, line 5: What about having a release variable? I think that's what you did for other uses of the rc flag, and that keeps things a bit more obvious for those use to thinking of release numbers especially when we get to GA?

ehelms:
ACK

katello-installer/katello-installer-base.spec
jlsherrill, line 5: as discussed the tag 2.4.0.rc1 was already in the git repo for katello installer, so best option was to go to rc1.1

PR #102, opened 2015-11-02
Fixes #12354 - obsolete uneeded gems

jlsherrill:


jlsherrill:
@ehelms  said he tried this and it didn't work, but he couldn't remember why, so retrying now

rubygem-katello/rubygem-katello.spec
```diff
@@ -140,6 +140,10 @@ BuildRequires: %{?scl_prefix}rubygem(qpid_messaging) < 0.31.0
 BuildRequires: %{?scl_prefix_ruby}rubygems-devel
 BuildRequires: %{?scl_prefix_ruby}ruby(rubygems)
 
+#old dependencies that are no longer needed, remove in katello 2.6
```
ehelms, line 143: 2.4

ehelms:
The test will come with upgrading with a scratch build that includes this - that's where it failed for me iirc.

rubygem-katello/rubygem-katello.spec
jlsherrill, line 143: No, i was implying to leave this obsoletes in there for two releases and then remove them.  I could see an argument for just one release though.

jlsherrill:
@ehelms this seems to work fine for me (updated to obsolete all 3 replaced rpms). 

@stbenjam since @ehelms is likely gone, are you okay with these changes?

stbenjam:
Looks ok to me, 2 releases sounds good to make sure we get nearly everyone.

ACK

stbenjam:
Also, hooray, HAML is :skull: dead

PR #104, opened 2015-11-10
rebuild of katello and installer for 2.4.0.rc2

jlsherrill:


stbenjam:
QU**ACK**

quack

stbenjam:
(That's an ACK to be clear)

PR #105, opened 2015-11-12
Add jenkins_api_client package

xprazak2:
This is required for foreman_pipeline

### DIFF
--- a/rubygem-jenkins_api_client/rubygem-jenkins_api_client.spec
+++ b/rubygem-jenkins_api_client/rubygem-jenkins_api_client.spec
@@ -1,0 +1,107 @@
+%{?scl:%scl_package rubygem-%{gem_name}}
+%{!?scl:%global pkg_name %{name}}
+%global gem_name jenkins_api_client
+
+Summary: Jenkins Api client with features focused on automating Job configuration programaticaly
+Name: %{?scl_prefix}rubygem-%{gem_name}
+Version: 1.4.1
+Release: 1%{?dist}
+Group: Developmanet/Languages
+License: MIT
+URL: https://github.com/arangamani/jenkins_api_client
+Source0: http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+Requires: %{?scl_prefix_ruby}ruby(rubygems)
+Requires: %{?scl_prefix_ruby}ruby
+
+%if "%{?scl_ruby}" == "ruby193" || (0%{?el6} && 0%{!?scl:1})
+Requires: %{?scl_prefix_ruby}ruby(abi)
+BuildRequires: %{?scl_prefix_ruby}ruby(abi)
+%else
+Requires: %{?scl_prefix_ruby}ruby(release)
+BuildRequires: %{?scl_prefix_ruby}ruby(release)
+%endif
+
+BuildRequires: %{?scl_prefix_ruby}rubygems-devel
+BuildRequires: %{?scl_prefix_ruby}ruby
+
+Requires: %{?scl_prefix_ruby}rubygem(json)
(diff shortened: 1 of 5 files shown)


ehelms:
@xprazak2 can you rebase this?

xprazak2:
@ehelms, I rebased.

ehelms:
Can you please add this package to the appropriate places in https://github.com/Katello/katello-packaging/blob/master/rel-eng/tito.props ?

xprazak2:
I added the package to the rel-eng/tito.props

rubygem-jenkins_api_client/rubygem-jenkins_api_client.spec
```diff
@@ -0,0 +1,106 @@
...
+Requires: %{?scl_prefix_ruby}rubygem(json)
+Requires: %{?scl_prefix_ruby}rubygem(mixlib-shellout) >= 1.1.0
+Requires: %{?scl_prefix_ruby}rubygem(terminal-table) >= 1.4.0
+Requires: %{?scl_prefix_ruby}rubygem(thor) >= 0.16.0
+Requires: %{?scl_prefix_ruby}rubygem(nokogiri) >= 1.6.0
+Requires: %{?scl_prefix_ruby}rubygem(nokogiri) < 2.0.0
+
+BuildRequires: %{?scl_prefix_ruby}rubygem(json)
+BuildRequires: %{?scl_prefix_ruby}rubygem(mixlib-shellout) >= 1.1.0
+BuildRequires: %{?scl_prefix_ruby}rubygem(terminal-table) >= 1.4.0
```
ehelms: mixlib-shellout and terminal-table will need to use `%{?scl_prefix}` since we build them into the `tfm-` namespace instead of the ruby scl namespace. Same is true in the Requires section above.

rubygem-jenkins_api_client/rubygem-jenkins_api_client.spec
```diff
@@ -0,0 +1,106 @@
...
+Requires: %{?scl_prefix}rubygem(thor) >= 0.16.0
+Requires: %{?scl_prefix}rubygem(nokogiri) >= 1.6.0
+Requires: %{?scl_prefix}rubygem(nokogiri) < 2.0.0
+
+BuildRequires: %{?scl_prefix}rubygem(json)
+BuildRequires: %{?scl_prefix}rubygem(mixlib-shellout) >= 1.1.0
+BuildRequires: %{?scl_prefix}rubygem(terminal-table) >= 1.4.0
+BuildRequires: %{?scl_prefix}rubygem(thor) >= 0.16.0
+BuildRequires: %{?scl_prefix}rubygem(nokogiri) >= 1.6.0
+BuildRequires: %{?scl_prefix}rubygem(nokogiri) < 2.0.0
```
ehelms, line 40: This is still going to fail since some of these packages are in the ruby193 SCL. I'll try to explain better. Currently we have two SCL's involved: ruby193 and tfm. ruby193 is the SCL where ruby, rails and some accompanying packages are kept and maintained outside of our control. tfm is a custom SCL maintained in foreman-packaging and to which we build any packages that we are maintaining whether they are plugins we developer or thirdparty gems. The macros here in the spec file represent each of those:

ruby193 = scl_prefix_ruby
tfm = scl_prefix

Thus, any package that is maintained in katello-packaging or foreman-packaging will be built with `tfm-` and thus uses `scl_prefix` to reference it in a spec file. All others come from `ruby193` and use `scl_prefix_ruby`.

ehelms:
ACK, thanks @xprazak2 ! Will build and re-run the pipeline PR

PR #106, opened 2015-11-12
Add mixlib-shellout, terminal-table and thor packages

xprazak2:
These are required for jenkins_api_client


ehelms:
I believe thor is already packaged in the SCL -- https://www.softwarecollections.org/repos/rhscl/ruby193/epel-7-x86_64/

rel-eng/tito.props
```diff
@@ -18,10 +18,16 @@ whitelist = katello-repos katello-agent
...
+whitelist = rubygem-mixlib-shellout
+  rubygem-terminal-table
+  rubygem-thor
 
 [katello-nightly-rhel7]
 disttag = .el7
 scl = tfm
+whitelist = rubygem-mixlib-shellout
+  rubygem-terminal-table
+  rubygem-thor
```
ehelms: Why do these need to be whitelisted here?

xprazak2:
@ehelms: I removed the whitelisting, it should get packaged as 'tfm' even without it, right? I believe I need thor in tfm as well...

ehelms:
As long Thor is available in an SCL that we use then we don't need to package it. Currently it is available in ruby193 and rh-ruby22.

xprazak2:
OK, I removed thor.

ehelms:
ACK - merging and will build and re-test the next one in line

PR #107, opened 2015-11-12
Add foreman_pipeline package

xprazak2:


ehelms:
Is foreman_pipeline Rails 4 compatible or is there a PR to update it to Rails 4? Since the packaging here will be moving to Rails 4 SCLs very soon, I wanted to check on that aspect.

xprazak2:
@ehelms: It is not Rails 4 compatible and there is no PR to update it yet. I guess I have work to do...

ehelms:
@xprazak2 I'd recommend that, otherwise we'd merge this and within a few days it would break

xprazak2:
@ehelms Thanks for warning, I will be back with an update

xprazak2:
I made changes to be ready for Rails 4, I hope I did not miss anything.

ehelms:
@xprazak2 double checking, the changes to foreman_pipeline are rails 4 and 3 compatible or did you have to make backwards incompatible changes?

xprazak2:
@ehelms the changes are backward compatible

rel-eng/tito.props
```diff
@@ -18,10 +18,12 @@ whitelist = katello-repos katello-agent
 [katello-nightly-rhel6]
 disttag = .el6
 scl = tfm
+whitelist = rubygem-foreman_pipeline
 
 [katello-nightly-rhel7]
 disttag = .el7
 scl = tfm
+whitelist = rubygem-foreman_pipeline
```
ehelms: These whitelists shouldn't be needed

ehelms:
I will work on getting this merged and built over the next day or two.

xprazak2:
Great! I removed the whitelisting from rel-eng/tito.props

ehelms:
Can you rebase and update this PR the same as you did the previous for the scl name changes and add the whitelisting?

rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
```diff
@@ -0,0 +1,124 @@
...
+Source0:    http://rubygems.org/downloads/%{gem_name}-%{version}.gem
+
+Requires:   %{?scl_prefix}rubygem(jenkins_api_client) < 2.0.0
+Requires:   %{?scl_prefix}rubygem(foreman_deployments) < 1.0.0
+Requires:   %{?scl_prefix}rubygem(net-ssh) >= 2.9
+Requires:   %{?scl_prefix}rubygem(net-ssh) < 3.0
+Requires:   %{?scl_prefix}rubygem(net-scp) >= 1.1
+Requires:   %{?scl_prefix}rubygem(net-scp) < 2.0
+Requires:   %{?scl_prefix}rubygem-katello >= 2.4.0
+Requires:   %{?scl_prefix}rubygem-katello < 3.0.0
```
ehelms, line 34: Looks like this needs to be a build requires as well.

xprazak2:
@ehelms, I added katello as build requires, but it is complaining about [missing packages](http://koji.katello.org/koji/getfile?taskID=384466&name=mock_output.log), requiring them in my spec does not seem to have an effect, any thoughts on what I am missing?

ehelms:
[test]

ehelms:
[test]

ehelms:
[test]

ehelms:
@xprazak2 made some progress, I think you will just need to update the spec based on the most recent error since it is erroring in the build phase. Check https://github.com/Katello/katello-packaging/blob/master/rubygem-katello/rubygem-katello.spec#L157-L158 for what you can do in yours

ehelms:
@xprazak2 new fialure during migration but seems code specific.

ehelms:
The current failure is because you are trying to include files in the RPM that don't exist in the gem itself. They would need to be included via https://github.com/theforeman/foreman_pipeline/blob/master/foreman_pipeline.gemspec#L16 and a new release of the gem put out. I would add the script/ directory there, but there isn't a need to deploy the Gemfile in this case since that is only used in development.

ehelms:
Will you update tito.props to build to katello-nightly-rhel6/7 instead of thirdparty?

xprazak2:
I updated tito.props, thanks for helping me with this @ehelms

ehelms:
ACK, thanks @xprazak2

PR #108, opened 2015-11-17
Fixes #12511: Remove rubygem-hooks as a dependency

ehelms:


ehelms:
This was removed as part of the scoped search conversion work

jlsherrill:
ACK

PR #114, opened 2015-11-25
Commit of package [katello-utils] minor release [2.4.0-4].

sean797:
PR Katello/katello-utils#2 has been merged so we need to bump the version number.


ehelms:
Looks like you did a tito tag here based on the update to the changelog which isn't needed for this update. All we need for the PR is to bump the version number in the spec file and reset the release to 1 (I'll do the tito tag and building once I merge the PR with those updates).

sean797:
Eric I'll leave this for you to do, if that's okay?

PR #115, opened 2015-11-30
Refs #12625 - include python-cherrypy as it was removed from epel7

jlsherrill:


stbenjam:
Based on the reason from https://admin.fedoraproject.org/pkgdb/package/python-cherrypy/, it's retired because it was included in EL 7.2.  Can't we just make that the minimum version required?

jlsherrill:
Centos 7.2 isn't released yet, so its not available.  This is a temporary fix until centos 7.2 is available.

stbenjam:
Ah ok, ACK from me

PR #117, opened 2015-12-04
Fixes #12708 - Collect the gutterball log files

bkearney:


ehelms:
Would you mind looking at https://github.com/Katello/foreman-gutterball/pull/10/files and resolving the differences? I would like to get this change in, but since foreman_gutterball controls gutterball installation I'd rather it also controlled this via the proposed inclusion script as in the referenced PR. This can either go into the rubygem as it is in the PR, or into this packaging repository and included as a source within rubygem-foreman_gutterball.

bkearney:
@ehelms that commit is better. If you can tie that pr to issue 12708 then this can be discarded.

ehelms:
Agreed in part -- it has been sitting for a while and could use active
development so we can n get it in.

PR #119, opened 2015-12-08
Fixes #12741: Move pulp/candlepin requires to metapackage

ehelms:


ehelms:
@cristifalcas if you have time to check whether this would solve your use case

This is also being done to solve the build issues with https://github.com/Katello/katello-packaging/pull/107 which will prevent us form having to tag into our buildroots all of the CP and Pulp packages for gems that depend on Katello.

cristifalcas:
Hi @ehelms, it looks like this fixes all my issues.

katello/katello.spec
```diff
@@ -28,15 +28,57 @@ BuildRequires: util-linux
...
 
+%if 0%{?rhel} == 6
+Requires: redhat-logos >= 60.0.14
+%endif
+
+Requires(post): chkconfig
+Requires(postun): initscripts coreutils sed
+Requires(pre): shadow-utils
+Requires(preun): chkconfig
+Requires(preun): initscripts
```
jlsherrill: i'm not sure these are actually needed anymore?  We aren't starting or stopping (or enabling) services, are we doing anything at all in %pre or %post in either rpm?

ehelms: Glad you asked that as I wondered the same thing myself -- I don't see any post or pre service related items given we no longer worry about deploying our own service or tasking anymore.

ehelms:
Updated to remove those post/pre requires

jlsherrill:
ACK

PR #121, opened 2015-12-11
Fixes #12782 Updating katello-backup

chris1984:
Updating katello-backup to capture /var/puppet/foreman_cache_data/


ehelms:
ACK - thanks @chris1984 !

PR #128, opened 2015-12-23
Remove rubygem-hashr

ehelms:


jlsherrill:
what was it that used this?

ehelms:
I can search back to find out, but its not in the Gemfile.lock so I removed it

jlsherrill:
oh actually it was tire.  ACK

PR #143, opened 2016-01-06
Include native extensions for qpid_messaging properly

ehelms:


jlsherrill:
ACK  /opt/theforeman/tfm/root/usr/lib64/gems/ruby/qpid_messaging-0.30.0/cqpid.so  appears

PR #144, opened 2016-01-11
fixes #13121 - Allows for incremental backups

cfouant:


ehelms:
@sean797 @evgeni  If either/both of you have time to overlook these changes given your past contributions in this area that would be greatly appreciated!

evgeni:
@ehelms sure thing, almost everything for better katello backups

@cfouant looks good from above, a few minor nitpicks  though:
- `-a` implies `-r`, so no need to pass that to `rsync`
- `-v` will create a lot of output (at least for `…/pulp`), it's pretty much useless in the log and will slow down the backup depending on the console/logging solution. better to add it only if `--verbose` was passed to `katello-backup`
- `-z` is not needed when we read/write to a local fs and we can save a few CPU cycles when omitting it
- you name the variables where the `--link-dest=…` param is `foo_dir`, that confused me at first because I did not see `--link-dest` in the final `rsync` call, would you mind renaming them to `foo_link_dest` or something?
- this change makes it impossible to make backups to some "special" filesystems (like vfat), not that we would care but this should be documented.

sean797:
I'm thinking tar with `--listed-incremental` may be better than using rsync? That way you can go back to any of the previous incremental backups, not just the latest one. Also it makes the backups that bit more portable than rsync. If we do use rsync were going to have to include the `--delete` flag as well.

Also we should check its being run as root, as otherwise some of the files in directory's like `/etc/pki/katello` will get permission denied errors.

komidore64:
@cfouant, it might be more helpful to have rsync's options in long-form here so that at a glance it's easier to figure out what it's doing.

sean797:
@cfouant As for the speed, you could backup everything but pulp, restart the services, display a message to the user saying not to do anything with content view's or sync any repos and/or set a lock on all the repos & content view versions until the backup finishes ? That should speed things up to an extent.

I'm not sure if this will just cause problems for users? if they decide it's taking too long and kill the backup process the locks on repos & content view version would remain.

evgeni:
@sean797 we run katello-backup in cron, so nobody would see that message here.

ideally, we would not backup pulp but ge-fetch the RPMs from CDN, but well... not implemented yet

ehelms:
I have not been following this closely. @evgeni @sean797 does the current state look good?

katello/katello-restore
```diff
@@ -26,7 +26,7 @@ def restore
   puts "Done.\n"
 
   puts "Resetting Katello"
-  `tar --selinux --overwrite -xzf config_files.tar.gz -C /`
+  `tar --selinux --overwrite --listed-incremental=config.snar -xzf config_files.tar.gz -C /`
```
sean797: There is no need to point `--listed-incremental` to a file during a restore, `/dev/null` is fine.

sean797:
I'll have a proper look early next week :+1:

evgeni:
I have no experience with tar's incremental archives, so I'd leave the actual review to @sean797 :)

katello/katello-restore
```diff
@@ -36,9 +36,9 @@ def restore
   puts "Done.\n"
 
   puts "Restoring backend data"
-  `tar --selinux --overwrite -xf pulp_data.tar -C /`
-  `tar --selinux --overwrite -xzf mongo_data.tar.gz -C /`
-  `tar --selinux --overwrite -xzf pgsql_data.tar.gz -C /`
+  `tar --selinux --overwrite --listed-incremental=pulp.snar -xf pulp_data.tar -C /`
```
sean797: Won't this create a `No such file or directory` error if `--skip-pulp-content` is used during the backup? It shouldn't be an issue but the user may not know that.

sean797:
2 nitpick but it works really well, nice job :+1:

katello/katello-restore
cfouant: @sean797 - Users should not be restoring from a backup where pulp content was skipped (outlined in the new documentation still awaiting merge). --skip-pulp option would only be utilized for things like debugging. I'll go ahead and fail the restore if the backup file is missing.

sean797:
Nice! ACK :+1:

ehelms:
Thanks @cfouant !!

lzap:
Sup folks! I am testing this for Satellite 6.2 and it works, but there is not a single word about restoring an incremental backup. I assume that I need to restore the full one and all the other incrementals, can you confirm? It is worth filing an issue to expand this topic in the help screen and in the manual as well.

lzap:
If that's the case, that might be lengthy process because we start and stop all services. For like 10 incrementals this can be bummer, we might want to extend the CLI with ability to read multiple directories or something like that. What you think?

sean797:
You just have to restore the multiple backups starting with the oldest one. Yes I agree we need to document it and getting katello-restore to read multiple directories sounds like a good idea to me, but we would need to make sure we clearly explain users need to ordered them correctly.

PR #147, opened 2016-01-22
Remove bootdisk and discovery since they aren't Rails 4 compatible

ehelms:


ehelms:
This is due to these lacking Rails 4 releases, and breaking repoclosure -- http://ci.theforeman.org/job/packaging_repoclosure/27892/console

jlsherrill:
ACK

PR #158, opened 2016-02-04
adding back candlepin-common

jlsherrill:
didnt realize that was a separate package


ehelms:
I feel like this should be accompanied with "psych!

ehelms:
ACK

PR #160, opened 2016-02-04
Update runcible to 1.6.0

parthaa:


rubygem-runcible/rubygem-runcible.spec
```diff
@@ -5,7 +5,7 @@
 
 Name:           %{?scl_prefix}rubygem-%{gem_name}
 Version:        1.5.0
-Release:        1%{?dist}
+Release:        2%{?dist}
```
ehelms: Should be 1 and the version needs to be 1.6.0

ehelms:
ACK - thanks @parthaa !

PR #161, opened 2016-02-06
Refs #13589 - replace gofer plugin config file

jlsherrill:
this config file is not meant to ever be customized by the user, but is needed
for the plugin to function


ehelms:
Do you want to go ahead and bump the release here too?

jlsherrill:
I figured we'd wait till release time to tag, pull in the new release, and bump the version in case there are other changes that come in.  I was more making this change so we didn't forget.

jlsherrill:
We'll also need a new updated version of gofer to support the feature in https://github.com/Katello/katello-agent/pull/34  

It should appear soon with pulp 2.8

ehelms:
Groovy - ACK

jlsherrill:
Need to add a dependency on gofer 2.7

jlsherrill:
2.7.4 to be exact

ehelms:
Looks like gofer 2.7.5 is staged to go into nightlies with the latest Pulp update if you want to fix this PR up and we can get it merged.

jlsherrill:
I figured i'd go ahead and bump the version of katello-agent.  Thoughts on what i should bump it to?  3.0 ?

ehelms:
Is this a backwards incompatible change?

jlsherrill:
It should not be afaik

ehelms:
Will this katello-agent still work with Katello 2.4 for example? If so, then I guess I don't see a reason for a major bump.

jlsherrill:
Well there could be  two questions that you are asking

a)  Will the new katello agent work with pulp 2.6 consumer code against pulp 2.8
No idea, we never test this.  If i had to wildly guess I would say yes.

b) Will the new katello agent work with pulp 2.6 consumer code against pulp 2.6
Yes, it should

EDIT: fixed question b)

ehelms:
So I guess no reason to bump to 3.0 unless we are just trying to keep katello-agent versioning the same as the main application?

jlsherrill:
@ehelms katello-agent 2.5.0 releaesd and this now uses that

ehelms:
This has the binary and not the git annexed file.

jlsherrill:
@ehelms updated

ehelms:
APJ

PR #166, opened 2016-02-12
add .ruby-version and .ruby-gemset to gitignore

komidore64:


ehelms:
ACK - thanks @komidore64 !

PR #165, opened 2016-02-11
Add OSTree packages to comps

parthaa:

PR #168, opened 2016-02-16
Fixes #13676 - Creating sub package for ostree

parthaa:


rubygem-katello/rubygem-katello.spec
```diff
@@ -95,6 +95,14 @@ Summary:    Documentation for rubygem-%{gem_name}
 %description doc
 This package contains documentation for rubygem-%{gem_name}.
 
+%package ostree
+Requires:   %{?scl_prefix}%{pkg_name} = %{version}-%{release}
+Summary:    Katello Ostree Plugin
+%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}-ostree}
```
ehelms: This line isn't needed, there is no ruby193-rubygem-katello-ostree

rubygem-katello/rubygem-katello.spec
```diff
@@ -95,6 +95,14 @@ Summary:    Documentation for rubygem-%{gem_name}
 %description doc
 This package contains documentation for rubygem-%{gem_name}.
 
+%package ostree
```
ehelms: I think we need to custom name this as RPM will try to use `-` to connect, and we want to use the more typical trend of `_`:

```
tfm-rubygem-katello-ostree
```

vs

```
tfm-rubygem-katello_ostree
```

rubygem-katello/rubygem-katello.spec
```diff
@@ -95,6 +95,14 @@ Summary:    Documentation for rubygem-%{gem_name}
...
+Requires:   %{?scl_prefix}%{pkg_name} = %{version}-%{release}
+Summary:    Katello Ostree Plugin
+%{?scl:Obsoletes: ruby193-rubygem-%{gem_name}-ostree}
+
+%description ostree
+This package contains documentation for ostree plugin for rubygem-%{gem_name}.
```
ehelms: Documentation for ostree plugin? I think it does more than that :)

rubygem-katello/rubygem-katello.spec
parthaa: do we do this anywhere else ?

parthaa: I can try to do something like `%package -n  %{?scl_prefix}rubygem-%{gem_name}_ostree`

ehelms: :+1: -- @jlsherrill do you agree?

jlsherrill: :+1: i think that makes sense

parthaa: @jlsherrill @ehelms updated

ehelms:
You'll want to add this to the comps as well.

parthaa:
> You'll want to add this to the comps as well.

Good call. Updated.

parthaa:
We need this guy in too https://github.com/Katello/katello-packaging/pull/165 ..

jlsherrill:
ACK from me, build looks good

ehelms:
:+1:

PR #170, opened 2016-02-18
Update rubygem-hammer_cli_katello to 0.0.22

komidore64:


beav:
I had some trouble with ./setup_sources.sh that was unrelated to this patch, but this patch seems reasonable.

ACK

ehelms:
@beav if you click through the testing that is run, it will try and setup sources and build this in koji which serves as a check that the thing in question was released properly upstream, can be resolved and the RPM build passes successfully.

ehelms:
ACK - thanks @komidore64 !

beav:
@ehelms cool, good to know :rainbow:

PR #181, opened 2016-03-04
fixes #14054 - ensure katello-agent service errors don't show during install

stbenjam:
The particular error mentioned in the bug is just a connection refused, goferd couldn't talk to the Katello, but the %post and %postun should always return 0 I suppose, even in goferd didn't work correctly.

Also pipe to /dev/null to avoid showing the traceback.

Thoughts?

### DIFF
--- a/katello-agent/katello-agent.spec
+++ b/katello-agent/katello-agent.spec
@@ -1,6 +1,6 @@
 Name: katello-agent
 Version: 2.4.0
-Release: 3%{?dist}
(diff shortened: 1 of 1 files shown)


ehelms:
Looks good to me -- @jlsherrill ?

jlsherrill:
ACK from me

ehelms:
Thanks @stbenjam !

stbenjam:
Thanks!

PR #182, opened 2016-03-08
Fixes #13199 - Remove pulp-nodes-parent as a requirement

johnpmitsch:


jlsherrill:
I think a pr needs to be opened against the puppet module:  https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L132  

as well otherwise the package will get reinstalled anytime the installer runs

ehelms:
Comps updates would be nice to e.g. https://github.com/Katello/katello-packaging/blob/af4c354f5bd567753485954e0d4f61d8b9471b52/comps/comps-katello-pulp-server-rhel6.xml#L26-L29

ehelms:
Looks like katello-debug is currently collecting some nodes related info too at the moment, I doubt that breaks the script but we should update it at some point.

johnpmitsch:
@ehelms @jlsherrill updated and opened up https://github.com/Katello/puppet-katello/pull/115 as well

comps/comps-katello-pulp-server-rhel6.xml
```diff
@@ -23,10 +23,6 @@
        <packagereq type="default">pulp-consumer-client</packagereq>
        <packagereq type="default">pulp-docker-plugins</packagereq>
        <packagereq type="default">pulp-docker-admin-extensions</packagereq>
-       <packagereq type="default">pulp-nodes-admin-extensions</packagereq>
-       <packagereq type="default">pulp-nodes-child</packagereq>
```
jlsherrill, line 27: I think this may break capsule installs due to https://github.com/Katello/puppet-pulp/blob/master/manifests/child/install.pp#L5

i'm not 100% of the best way to do it because if katello doesn't use pulp::child class there lots of other things aren't configured properly

ehelms, line 27: If we install a full pulp vs pulp::child what gets configured wrong? As in, can we identify that easily to correct or is this 3.1 technical debt?

jlsherrill, line 27: I don't honestly know as we don't install a full (non-child) pulp currently today iirc

ehelms:
@jlsherrill are you OK with this change now that the installer work is in?

jlsherrill:
Yep, ACK thanks @johnpmitsch

PR #183, opened 2016-03-09
Fixes #14125 - Removing gutterball from script

chris1984:
removing /etc/gutterball from script to prevent keystore issues upon reinstall.


chris1984:
Going to make a new PR and redmine for this issue and close this out. I am seeing multiple things not removed from katello-remove so I will just make 1 major commit for that.

PR #192, opened 2016-03-16
Fixes #14189: Move repo definitions to be dist based 

ehelms:
We were using subdirectories such as RHEL/7 and RHEL/7Server to
denote locations for different distributions. This moves to a dist
based approach (e.g. el7) to support more platforms such as 7Server,
Centos or something like 7Workstation.

### DIFF
--- a/repos/katello-client.repo
+++ b/repos/katello-client.repo
@@ -2,7 +2,7 @@
 
 [katello-client]
 name=Katello Client Nightly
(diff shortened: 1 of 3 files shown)


ehelms:
You can see this reflected in Koji staging: http://koji.katello.org/releases/yum/katello-nightly/client/

ehelms:
Also an update to kdeploy for this change https://github.com/Katello/katello-deploy/pull/207

jlsherrill:
ACK

PR #194, opened 2016-03-18
Fixes #14260: Ensure the leading dot is removed from dist in repos RPM

ehelms:


repos/katello-repos.spec
```diff
@@ -46,7 +46,8 @@ install -m 644 %{SOURCE1} %{buildroot}%{_sysconfdir}/yum.repos.d/
 install -m 644 %{SOURCE2} %{buildroot}%{_sysconfdir}/pki/rpm-gpg/RPM-GPG-KEY-katello
 
 for repofile in %{buildroot}%{_sysconfdir}/yum.repos.d/*.repo; do
-    sed -i 's/@DIST@/%{dist}/' $repofile
+    trimmed_dist=`echo %{dist} | sed 's/.//'`
```
thomasmckay: This will trim the first character always. 's/.//' would just strip '.'. Not sure if that matters here or not.

jlsherrill:
APJ

repos/katello-repos.spec
thomasmckay: Sorry, should be 's/^.//' if just wish to trim leading '.'.

ehelms: I am confused -- we want to strip just the dot off the front.

thomasmckay:
```
[vagrant@katello foreman]$ echo "abc" | sed 's/.//'
bc
[vagrant@katello foreman]$ echo "abc" | sed 's/\.//'
abc
[vagrant@katello foreman]$ echo "a.bc" | sed 's/\.//'
abc
[vagrant@katello foreman]$ echo "a.bc" | sed 's/^\.//'
a.bc
[vagrant@katello foreman]$ echo ".a.bc" | sed 's/^\.//'
a.bc
```

I'm saying that this is the correct sed:

```
sed 's/^\.//'
```

thomasmckay: Sorry, I see my earlier comments lost a slash in the comment rendering, hence the confusion.

ehelms:
Updated per @thomasmckay suggestion

thomasmckay:
APJ

thomasmckay:
APJ if don't care what first character being removed is

ehelms:
I updated per IRC discussion to remove just the first explicit dot.

PR #199, opened 2016-04-07
Fixes #14320 - packaging for client bootstrap script

mccun934:


mccun934:
@ehelms I'm not following what I need todo with git annex for that build failure. I committed the source and pushed it with this PR..

ehelms:
If you run the `setup_sources.sh` script you should see what I mean. The source from the spec, in this case `https://codeload.github.com/Katello/%{name}/tar.gz/%{version}` needs to match the filename that is in git annex. So if you `wget https://codeload.github.com/Katello/client-bootstrap/tar.gz/1.0.0` for example, it will get a file named '1.0.0'. That is what needs added to git annex.

comps/comps-katello-server-rhel6.xml
```diff
@@ -29,6 +29,7 @@
        <packagereq type="default">tfm-rubygem-foreman_pipeline</packagereq>
        <packagereq type="default">tfm-rubygem-hammer_cli_katello</packagereq>
        <packagereq type="default">tfm-rubygem-hammer_cli_import</packagereq>
+       <packagereq type="default">client-boostrap</packagereq>
```
ehelms: This is intended for the server repositories and not the client repositories?

mccun934: yes, it is going to land in /var/lib/www/html/pub and clients will wget it, similar to the consumer-certs RPM. We need it accessible prior to registration in an unprotected public location so you can 'bootstrap' and get registered and then get content.

ehelms:
Three thoughts:

1) Should we also provide this in the client repos directly?
2) Which RPM do you envision requiring this so it gets installed? Or do you see the installer doing that?
3) I assume we'd want this to be available on a Capsule too? For the case of isolation given the certs bootstrap RPM lives on the Capsules today?

mccun934:
1) we could, but would require some specfile work to have it sometimes land in /var/www/html/pub and sometimes land in /usr/sbin/bootstrap.py. I suppose we could have subpackages and you install client-bootstrap-server to put it in var/www and client-bootstrap-client ;) to put it in /usr/sbin

To be clear, the usecase is, on a pre-existing client that you want to get puppet + sub-man configured and managed in Foreman you do:

```
wget http://katello.example.com/pub/bootstrap.py
python bootstrap.py -s katello.example.com ...
```

If we had subpackages as you outline above we could do:

```
rpm -Uvh http://fedorapeople.org/groups/katello/releases/yum/3.0/client/RHEL/5Server/x86_64/client-bootstrap-latest.rpm
python bootstrap.py -s katello.example.com ...
```

The above scenario would be convenient but not sure if we need it right now.

2) Installed on the server along with katello. I'll be doing a separate PR to the katello spec to pull this in with a 'yum install katello' but I wanted this to exist by itself first.

3) Yes, just like the certs RPMs.

ehelms:
For #3, including it on the Katello meta RPM will work for the server but not for the capsules where we do not have a specific capsule meta RPM. We currently just have the user install the installer. Which means we could either:

 1) attach it as a requires of the installer
 2) set it as an installed package by puppet-capsule and let the installer run itself ensure the RPM is installed

mccun934:
I'll do (2), seems wrong to have hanging off as a dep of the installer since the installer itself doesn't need it.

mccun934:
will this get both the Katello server and the Capsule:

https://github.com/Katello/puppet-capsule/pull/81

ehelms:
Yep - it should. Oh, that reminds me of one other question I had, 'client-bootstrap' is fairly generic name if you are looking at installed or available RPMs. Should we "namespace" this at all?

mccun934:
I'll rename it in the specfile to **katello-client-bootstrap**

ehelms:
Can you rename the directory too? Helps when the specfile and directory matchup

mccun934:
@ehelms done .. renamed the source repo too.

ehelms:
ACK - I'll merge this, add it to our koji tags and build it.

PR #203, opened 2016-04-13
Fixes #14608 - avoid collection of squid with sosreport

lzap:
It's in https://github.com/sosreport/sos/blob/master/sos/plugins/squid.py


ehelms:
ACK - thanks @lzap

PR #206, opened 2016-04-18
Fixes #14673: Obsolete pulp-nodes* packages to resolve upgrade

ehelms:


jlsherrill:
I wonder if we'll have the same issue on the capsules?  

ACK

ehelms:
Most likely it will, I guess we could attach an obsolete to the installer package? We currently use the same installer package on the server and capsule thus if we moved it there we would not need this change.

jlsherrill:
Lets do that then, we'd have to add 'pulp-nodes-child' as well.

ehelms:
Moved to the installer package

jlsherrill:
ACK

PR #209, opened 2016-04-22
fixes #14784 - fix katello-backup syntax error

narbutas:
Fixed an error while trying to create Katello backup in 3.0 version

/usr/bin/katello-backup:115: syntax error, unexpected end-of-input, expecting keyword_end

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -20,7 +20,7 @@ optparse = OptionParser.new do |opts|
     else
       opts.abort("Previous backup directory does not exist: #{dir_path}")
     end
-
+  end
   opts.on("--online-backup", "Backup Pulp data online") do |online|
     @options[:online] = online
   end
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -21,6 +21,7 @@ optparse = OptionParser.new do |opts|
       opts.abort("Previous backup directory does not exist: #{dir_path}")
     end
   end
+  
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -20,7 +20,7 @@ optparse = OptionParser.new do |opts|
     else
       opts.abort("Previous backup directory does not exist: #{dir_path}")
     end
-
+  end
```
stbenjam, line 23: It'd look better if there was a blank line after this so the two blocks are separated.

narbutas, line 23: Thank you for noticing that, added extra empty line.

stbenjam:
Thanks, looks good to me, @cfouant should take a look though as she knows the most about this script.

I think we probably need a redmine issue too - we don't enforce that in this repo, but having one will make sure this gets tracked as added to Katello 3.0.

http://projects.theforeman.org/projects/katello/issues/new

narbutas:
Thank you for information. Added bug report: http://projects.theforeman.org/issues/14784

stbenjam:
[test]

stbenjam:
Ah she's out today and it's just a small change so merging.  ACK

Thanks @narbutas

stbenjam:
Merged into master as ec5d38b718e71ca6649fce1152df3820dee23b02 as well

PR #210, opened 2016-04-26
rebuild for ror42

jlsherrill:


ehelms:
FYI this will only try and build the first 5, we can change this upper limit just hadn't need a reason to before (was put in place to prevent overload from mass changes).

beav:
APJ

jlsherrill:
trimmed to 5 builds and opened a 2nd pr for the rest non-rails engine updates

jlsherrill:
@beav mind a quick re-review?

beav:
@jlsherrill ACK lgtm

PR #213, opened 2016-04-27
fixes #14858 - removes gutterball

cfouant:
This will remove gutterball from all existing installs, and removes install of gutterball for new Satellites.

Associated PRs:
http://github.com/Katello/puppet-certs/pull/85
http://github.com/Katello/puppet-katello/pull/124
http://github.com/Katello/puppet-katello_devel/pull/74
http://github.com/Katello/katello-installer/pull/329


stbenjam:
I get an install without gutterball, but trying the upgrade steps on a katello with gutterball I run into this:

```
[root@hi-gutterball ~]# rpm -e gutterball tfm-rubygem-foreman_gutterball tfm-rubygem-hammer_cli_gutterball
error: Failed dependencies:
    tfm-rubygem-foreman_gutterball is needed by (installed) katello-common-3.1.0-3.nightly.el7.noarch
    tfm-rubygem-hammer_cli_gutterball is needed by (installed) katello-common-3.1.0-3.nightly.el7.noarch
```

I think you'll need to remove those dependencies: https://github.com/Katello/katello-packaging/blob/master/katello/katello.spec#L117-L118

cfouant:
Thanks @stbenjam  - I also removed mention of gutterball files from katello-remove... not sure if we want to keep that in or not? As in, will it be ensured that everybody utilizing this updated katello-remove will have already had gutterball removed?

stbenjam:
> As in, will it be ensured that everybody utilizing this updated katello-remove will have already had gutterball removed?

Hm, good question, I'm not sure :) I think if they have this version of katello-remove that means they will have already run the upgrade steps, so I _think_ we can probably remove the gutterball from katello-remove too.

stbenjam:
Ah I see you already removed it from katello-remove too, missed that.  Looks good.  Other than @ehelms  comments on the installer PR, I think this is just about ready to go.  

I"ll give it all another try tomorrow to make sure we didn't miss anything.

stbenjam:
ACK

PR #214, opened 2016-04-29
Update foreman_pipeline to 0.0.11

xprazak2:
New release with updated dependencies.


xprazak2:
Seems like build succeeds, but my assets cannot be found when the package actually installs. Will do some digging.

xprazak2:
Issue with assets resolved, everything seems to be ok now.

rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
```diff
@@ -11,54 +11,39 @@
 
 %global gem_name foreman_pipeline
 %global foreman_dir /usr/share/foreman
-%global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
+%global foreman_bundlerd_dir %{foreman_dir}/bundler.d
```
ehelms: This shouldn't be needed as the macro handles it. Do you get errors if you don't define it here?

rubygem-foreman_pipeline/rubygem-foreman_pipeline.spec
```diff
@@ -11,54 +11,39 @@
...
 
-%define rubyabi 1.9.1
+%global scl_rake /usr/bin/%{?scl:%{scl_prefix}}rake
+%global scl_ruby_bin /usr/bin/%{scl:%{scl_prefix}}ruby
```
ehelms: This shouldn't be needed as the macro handles it. Do you get errors if you don't define it here?

xprazak2:
I removed `%global` definitions for rake, bundler.d and ruby, it seems be ok.

ehelms:
Thanks @xprazak2 ! I will build it into nightly and 3.0

PR #215, opened 2016-04-29
Fixes #14881 - Add in /etc/qpid-dispatch/qdrouterd.conf for files whi…

bkearney:
…ch are collected


ehelms:
ACK - thanks @bkearney !

PR #218, opened 2016-05-11
fixes #15010 - fix katello-service man page

stbenjam:
Man page currently ends up in the wrong format and in
/usr/share/man/man8/katello-service.8.asciidoc.gz


ehelms:
ACK - thanks @stbenjam !

PR #221, opened 2016-05-14
Fixes #15051 - Update katello-remove with help message

chris1984:
Added a --help flag to katello-remove to let the user know what this script does.

### DIFF
--- a/katello/katello-remove
+++ b/katello/katello-remove
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+for var in $@; do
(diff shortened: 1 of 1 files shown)


stbenjam:
Can we catch `-h` too?

chris1984:
@stbenjam ack, good catch. Fixed to catch -h and delivered with a squashed commit. If it looks good can I get a merge? I want to send in the 2nd commit for the other piece of katello-remove which would cause a merge conflict.

stbenjam:
ACK! Thanks @chris1984

PR #229, opened 2016-05-20
Fixes #15146 - service-wait now correctly handles tomcat

parthaa:


katello/service-wait
```diff
@@ -57,7 +57,7 @@ before_start() {
 # after start or restart
 after_start() {
   case "$SERVICE" in
-    tomcat6|tomcat7)
+    tomcat|tomcat6|tomcat7)
```
stbenjam: I think tomcat7 can actually be removed, not sure that's the name of any service

ehelms: A Redmine issue to pull this in would be nice as well.

bbuckingham: Should 'before_stop' also be updated?

ehelms:
Thanks @parthaa !

PR #230, opened 2016-05-30
Fixes #15230 - Require newer katello-service

ares:
Otherwise upgrading of older version fails

```
Upgrading...
Upgrade Step: stop_services...
Upgrade Step: start_databases...
Upgrade step start_databases failed. Check logs for more information.
```

since it uses `katello-service start --only mongod,postgresql`, older version does not have the support for `--only`


ehelms:
I think the change is fine but reading through the issue, we tell users to 'yum update' which covers this issue.

PR #233, opened 2016-06-06
Fixes #15299 - preserve legacy config during upgrade

mbacovsky:
The legacy config files are renamed to `.rpmsave` during yum update and migration script is unable to locate them. This patch is renaming the files back before migration.

I also moved calling installer migrations to `%posttrans` which is more practical and allows other packages to extend the migrations.

### DIFF
--- a/katello-installer/katello-installer-base.spec
+++ b/katello-installer/katello-installer-base.spec
@@ -32,7 +32,11 @@ Requires: katello-service >= 3.0.0
 %description -n foreman-installer-katello
 A set of tools for installation of Katello and and Capsule.
 
-%post -n foreman-installer-katello
+%posttrans -n foreman-installer-katello
+mv /etc/katello-installer/katello-installer.yaml.rpmsave /etc/katello-installer/katello-installer.yaml >/dev/null 2>&1 || :
+mv /etc/capsule-installer/capsule-installer.yaml.rpmsave /etc/capsule-installer/capsule-installer.yaml >/dev/null 2>&1 || :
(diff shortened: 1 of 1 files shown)


mbacovsky:
Scratch build: http://koji.katello.org/koji/taskinfo?taskID=476214

beav:
Can `%config(noreplace)` be removed instead?

Unfortunately it may take two runs for this to work. The RPM db uses the file's flags that are defined from the prior install, so it takes _two_ runs to clear a `noreplace`.

mbacovsky:
@beav, after considering this I'm not sure this is what we want. We are removing the file completely and I don't see a way to do it in two runs safely. Do you have an idea how to do that?

I was also thinking about migrating the content of the .rpmsave instead, but I decided not to pollute migrations with packaging specific workarounds (which would be pointless on e.g. Foreman installation on Debian).

beav:
@mbacovsky on a pre-upgrade installation, which package owns `/etc/katello-installer/katello-installer.yaml`? I don't have a pre-upgrade machine to check at the moment.

I'm not against doing this in a migration, I don't know if RPM will let us do what we need done, and I would rather have code in an migration than in spec file.

mbacovsky:
@beav, In Katello 2.4 the the config files are owned by `katello-installer` and `capsule-installer` respectively.

My goal in the new installer is to check if there is old config available and transform it to the new format as soon as possible. Installer shouldn't care how the file is managed, it should be up to the user if she controls the config file with rpm, apt, puppet or manually and it should be transparent.

ehelms:
@beav please to re-review

beav:
ACK, tested and works

PR #234, opened 2016-06-09
fixes #15366 - add yum-plugin-security dep to el5/6 packages

komidore64:


katello-agent/katello-agent.spec
```diff
@@ -24,6 +24,9 @@ Requires: python-gofer-proton >= 2.5
 Requires: python-pulp-agent-lib >= 2.6
 Requires: pulp-rpm-handlers >= 2.6
 Requires: subscription-manager
+%if 0%{?rhel} < 6
```
ehelms: This would only install it on EL5.

komidore64: doh! that should be a `7`!

ehelms:
Can you give a brief comment on why this fixes the described issue?

komidore64:
This spec change isn't actually the fix. The fix is in `pulp-2.8.4` and related packages. Also requires `yum-plugin-security` to be included explicitly on RHEL versions older than `7` because the yum on those OSes hadn't slurped in `yum-plugin-security` into core yet.

ehelms:
ACK - sorry for the long delay!

komidore64:
:+1:

PR #242, opened 2016-06-22
building katello-agent 2.6.0

jlsherrill:


jlsherrill:
Will be pulled in to 3.1

katello-agent/katello-agent.spec
```diff
@@ -1,5 +1,5 @@
 Name: katello-agent
-Version: 2.5.0
+Version: 2.6.0
 Release: 3%{?dist}
```
stbenjam: Should this go back down to 1?

jlsherrill: Thanks!  50% of the time I remember this everytime

stbenjam:
ACK

PR #243, opened 2016-06-23
updates for katello 3.1

jlsherrill:


katello/katello.spec
```diff
@@ -6,7 +6,7 @@
 
 Name:       katello
 Version:    3.1.0
-Release:    3.nightly%{?dist}
+Release:    4.rc1.%{?dist}
```
ehelms: The dot after rc1 shouldn't be needed

stbenjam:
ACK

PR #250, opened 2016-07-11
Change repos description form nightly to 3.1

ehelms:


jlsherrill:
APJ

stbenjam:
I did this too on the release I did, how about a sed in the %build so we don't have to change the repo files every time?

Something crazy like this (untested, and needs a ruby BuildRequires):

```
REPO_VERSION=$(echo "puts '%{release}' =~ /nightly/ ? 'Nightly' : '%{version}'.split('.')[0..1].join('.')" | ruby)
sed -ie "s/@VERSION@/$REPO_VERSION/" $repofile
```

jlsherrill:
@stbenjam opened a PR for that in master here:  https://github.com/Katello/katello-packaging/pull/256  (in addition made it so we didn't have to change the urls every time too)

will probably just merge this and not worry about backporting #256

stbenjam:
sure ACK

PR #256, opened 2016-07-22
allow for easier repo version setting

jlsherrill:


repos/katello-repos.spec
```diff
@@ -1,3 +1,6 @@
+%global repo_version nightly
+%global repo_name    Nightly
```
stbenjam: I guess this still needs updating every release? We have the info below in the Version and Release, would my suggestion on the other PR not work?

```
REPO_VERSION=$(echo "puts '%{release}' =~ /nightly/ ? 'Nightly' : '%{version}'.split('.')[0..1].join('.')" | ruby)
sed -ie "s/@VERSION@/$REPO_VERSION/" $repofile
```

jlsherrill: oh right!  will adapt

jlsherrill: @stbenjam updated!

stbenjam:
Awesome! ACK

PR #257, opened 2016-07-25
Fixes #15823 - Do not send mail on cron error

chris1984:
Fixing cron task so it does not send email to stdout


akofink:
LGTM - ACK

ehelms:
Thanks @chris1984 !

PR #259, opened 2016-07-26
Fixes #15845 - add cronjob to clean puppet envs

jlsherrill:
and add a new capsule subpackage to katello to install it on capsules

### DIFF
--- a/comps/comps-katello-server-rhel6.xml
+++ b/comps/comps-katello-server-rhel6.xml
@@ -13,6 +13,7 @@
     <packagelist>
        <!-- katello core packages -->
        <packagereq type="default">katello</packagereq>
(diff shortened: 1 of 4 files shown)


ehelms:
Where are you planning to include this for installation?

jlsherrill:
@ehelms was thinking of updating the capsule install and upgrade docs.  Any other thoughts?

ehelms:
ACK -- @jlsherrill planning a follow up PR that has the katello-capsule RPM require the installer so we can instruct users to just install this package?

jlsherrill:
pulp issue filed:  https://pulp.plan.io/issues/2108

Will do followup PR

PR #260, opened 2016-07-28
refs #15938 - install parser_cache from tarball

komidore64:
depends on katello/katello-installer#382.

### DIFF
--- a/katello-installer/katello-installer-base.spec
+++ b/katello-installer/katello-installer-base.spec
@@ -106,7 +106,7 @@ install -d -m0755 %{buildroot}/%{_datadir}/foreman-installer-katello-devel
 
 install -d -m0755 %{buildroot}/%{_sbindir}
 
-cp -dpR checks modules hooks %{buildroot}/%{_datadir}/katello-installer-base
+cp -dpR checks modules hooks parser_cache %{buildroot}/%{_datadir}/katello-installer-base
 
 cp -dpR bin/capsule-certs-generate %{buildroot}/%{_datadir}/foreman-installer-katello/bin/capsule-certs-generate
 cp -dpR bin/katello-certs-check %{buildroot}/%{_datadir}/foreman-installer-katello/bin/katello-certs-check
@@ -132,6 +132,7 @@ ln -sf %{_datadir}/foreman-installer-katello/bin/capsule-remove %{buildroot}/%{_
 %{_datadir}/katello-installer-base/modules
 %{_datadir}/katello-installer-base/hooks
 %{_datadir}/katello-installer-base/checks
+%{_datadir}/katello-installer-base/parser_cache
 %doc README.*
 
 %changelog


katello-installer/katello-installer-base.spec
```diff
@@ -16,16 +16,22 @@ Source0: https://github.com/Katello/katello-installer/archive/%{version}%{?preve
...
 Requires: katello-selinux
 Requires: openssl
 Requires: katello-certs-tools
 Requires: foreman-proxy
 
+BuildRequires: %{?scl_prefix}puppet >= 3.4.0
+BuildRequires: %{?scl_prefix}rubygem-kafo >= 0.8.0
+%if 0%{?fedora} >= 22
+BuildRequires: %{?scl_prefix}rubygem-puppet-strings
+%endif
```
ehelms: Since we don't build for Fedora, I suppose this requirement isn't needed in our case?

katello-installer/katello-installer-base.spec
```diff
@@ -99,6 +105,9 @@ sed -ri 'sX\:installer_dir.*$X:installer_dir: %{_datadir}/katello-installer-base
 sed -ri 'sX\:module_dirs.*$X:module_dirs: \["%{_datadir}/katello-installer-base/modules", "%{_datadir}/foreman-installer/modules"\]Xg' config/*.yaml
 sed -ri 'sX\:hook_dirs.*$X:hook_dirs: \["%{_datadir}/katello-installer-base/hooks"\]Xg' config/*.yaml
 
+# generate parser-cache
+kafo-export-params -c config/katello.yaml -f parsercache --no-parser-cache -o %{_datadir}/foreman-installer/katello.yaml
```
ehelms: I notice the foreman-installer is doing this step during the generation of the source tarball and bundling it in. Why do we want to do it in the RPM?

komidore64: Mostly because I'm flying blind and just trying to make it work for right now :) Totally not opposed to moving the cache generation to somewhere else.

From what I can tell, it looks like `foreman-installer` is [generating the parser cache](https://github.com/theforeman/foreman-installer/blob/23235877163bde6a2744a791fc20d483d3cb9570/Rakefile#L44-L46) during the RPM's [`%build`](https://github.com/theforeman/foreman-packaging/blob/306312d519f1f23aec596c0307ec047d8ecabaa9/foreman-installer/foreman-installer.spec#L44-L55) phase. Have I missed a step somewhere?

katello-installer/katello-installer-base.spec
komidore64: Hah, oh right. Good point. Would `0%{?rhel} >= 7` be better?

katello-installer/katello-installer-base.spec
stbenjam: We have a similar rake task in our katello-installer so I would think it should be added there, the only difference is they run the rake build in the RPM spec, and we do it out of band somewhere else (jenkins?)

ehelms: By rake build do you mean rake pkg:generate_source ? They run that to
generate source in Jenkins the same way for nighties afaik.

stbenjam: Oh  I see, they use 2 rake tasks, they use the generate source and then this separate build one in %build.  I guess to keep all of the build implementation details in the Rakefile

komidore64: So where should I move this to?

[`rake pkg:generate_source`](https://github.com/Katello/katello-installer/blob/master/Rakefile#L52-L65)?

stbenjam: However you want, but I'd prefer to do it the same way as Foreman, which would be to use the rake build task

ehelms:
[test]

ehelms:
[test]

ehelms:
ACK - thanks @komidore64

PR #267, opened 2016-08-09
Fixes #16028 - Ignoring proxy when pinging services

daviddavis:
Same change basically as https://git.io/v6csS. During upgrades, [the installer calls this line to ping tomcat](https://git.io/v6ccw) and we're seeing upgrades fail if `$https_proxy` is set.


ehelms:
ACK - thanks @daviddavis !

PR #269, opened 2016-08-11
Refs #13017 - cleanup qpid from comps

jlsherrill:
some qpid packages are still required, but since
they are only dispatch router related moving those
to katello repo


ehelms:
ACK

PR #275, opened 2016-08-29
fixes #16373 - include rex services in katello-service

stbenjam:
Katello service checks to see if a service is existent before it tries
to do anything with it, so I think it's worthwhile to include
smart_proxy_dynflow_core in it's list.


iNecas:
:+1:

iNecas:
:+1:

sean797:
Can you update the bash completion script as well? https://github.com/Katello/katello-packaging/blob/master/katello/katello-service-bash_completion.sh#L4

stbenjam:
Updated, thanks

ehelms:
ACK

PR #281, opened 2016-09-01
 Make sure katello-backup does not override permissions for existing directories

tchellomello:
This patch fixes katello-backup to avoid to override permissions from directories that already exists on the system. 

For instance, if the user by mistake type `katello-backup /tmp`, the `/tmp` permissions will override to 0770 causing a lot of issues on the system. 

``` bash
# ./katello-backup /tmp
**** ERROR: Directory already exists. Please specify a new export directory ****
Usage: katello-backup /path/to/dir [options]
 eg: $ katello-backup /tmp/katello-backup
        --skip-pulp-content          Create backup without Pulp content
        --incremental [PREVIOUS_BACKUP_DIR]
                                     Backup changes since previous backup
        --online-backup              Keep services online during backup
```

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -42,6 +42,12 @@ if @dir.nil?
   puts "**** ERROR: Please specify an export directory ****"
   puts optparse
   exit(-1)
+end
+
+if Dir.exist?(@dir)
+  puts "**** ERROR: Directory already exists. Please specify a new export directory ****"
+  puts optparse
+  exit(-1)
 else
   @time = Time.now
   puts "Starting backup: #{@time}"


ehelms:
[test]

cfouant:
NACK - I believe many customers currently use the same directory for backup every time, so not allowing a directory that already exists is not a viable solution.

stbenjam:
Personally I'd expect running `katello-backup /tmp` to do something like creating `/tmp/katello-backup-$TIMESTAMP` and then I think that'd fix @tchellomello's concern.

Dumping directly into the directory and changing it's permissions (like in the /tmp case, which wipes out the sticky bit) has really bad consequences

cfouant:
@stbenjam - That actually is how I originally had it, however it got NACK'ed with commentary that the user should be able to define their own backup directory. I'll bring it back up with the upper echelons.

tchellomello:
@cfouant this happened with a user back in the day. they ran the script on the /tmp and caused the system to break. we should handle this better.

cfouant:
I've received approval to create a subdirectory from user defined directory which will solve this issue (amongst others :+1:  !)

tchellomello:
@cfouant  thanks Since you are working on this, I'll close this PR.

PR #285, opened 2016-09-12
refs #16523 - fix error in find command

stbenjam:


beav:
@stbenjam that is funny, I eyeball-checked specifically that the P3 and P4 removals were the same, but still missed this as well. `me--`

tested both new commands, ACK

stbenjam:
Yup, must have have been something like this: https://www.reddit.com/r/explainlikeimfive/comments/1ruera/why_after_reading_this_does_the_the_brain_not/

Thanks!

PR #287, opened 2016-09-15
Update rubygem-hammer_cli_import to 0.11.3

ehelms:


stbenjam (changes requested):
Please update the RPM spec to match the gemspec in terms of dependencies, that's why this happened to begin with.

If the deps were right, repoclosure would've failed when hammer_cli_katello was released and we'd have known sooner

ehelms:
@stbenjam updated the spec file

rubygem-hammer_cli_import/rubygem-hammer_cli_import.spec
```diff
@@ -25,7 +25,8 @@ Obsoletes: rubygem-%{gem_name} < 0.10.21-3
 Requires: %{?scl_prefix_ruby}ruby(release)
 Requires: %{?scl_prefix_ruby}ruby(rubygems)
 Requires: %{?scl_prefix}rubygem(hammer_cli) >= 0.1.3
```
stbenjam: There's no minimum in the gemspec

stbenjam (changes requested):

rubygem-hammer_cli_import/rubygem-hammer_cli_import.spec
ehelms: Good call - updated

stbenjam (approved):

stbenjam:
ack!

PR #288, opened 2016-09-16
bumping to 3.2 rc1

johnpmitsch:


jlsherrill (approved):
ACK

repos/katello-repos.spec
```diff
@@ -17,6 +17,9 @@ BuildArch:      noarch
 BuildRequires: sed
 BuildRequires: ruby
 BuildRequires: python
+%if 0%{?fedora} > 23
+BuildRequires: rubygems
+%endif
```
ehelms, line 22: Why is this required for the katello-repos RPM?

katello/katello.spec
```diff
@@ -6,7 +6,7 @@
 
 Name:       katello
 Version:    3.2.0
-Release:    2.nightly%{?dist}
+Release:    2.rc1%{?dist}
```
ehelms, line 9: Should probably bump the release number here to be safe.

ehelms (commented):
I know this got merged already, but worth considering for updates.

repos/katello-repos.spec
johnpmitsch: fedora 24 added a rubypick package (ruby version manager) that is required to be installed in order to run the ruby command. Rubypick by itself isn't sufficient and it requires rubygems to run (not sure why on that part).

PR #294, opened 2016-09-20
Fixes #16626: Properly replace INSTALLER_DIR in capsule-certs-generate

ehelms:
This was causing the capsule-certs-generate to not have the path
to the installed katello-installer modules for the parser cache
path.


johnpmitsch:
ACK

PR #295, opened 2016-09-22
fixes #16655 - remove clean puppet envs script

stbenjam:
This can remove directory environments that while empty have hosts assigned to them in Foreman, breaking clients and generating error reports in Foreman.

What do we do about the katello-capsule package? Obsolete it?


stbenjam:
After talking to @jlsherrill: Nevermind, with `-maxdepth 0` the bug won't occur.  
- Empty, but potentially still valid puppet environments have a `modules` dir
- Environments from deleted content views do not.

PR #297, opened 2016-09-24
Fixes #17114 - foreman-debug counts size instead lines

pmoravec:
.. instead of tailing all (log)files

Update calling add_files (modified in foreman-debug).

Signed-off-by: Pavel Moravec pmoravec@redhat.com

### DIFF
--- a/katello/katello-debug.sh
+++ b/katello/katello-debug.sh
@@ -30,8 +30,8 @@ copy_files() {
 
 
 # Installer
-add_files /var/log/{katello,capsule}-installer/*
-add_files /etc/{katello,capsule}-installer/*
+add_files /var/log/foreman-installer/*
+add_files /etc/foreman-installer/*
 add_cmd "find /root/ssl-build -ls | sort -k 11" "katello_ssl_build_dir"
 add_cmd "find /etc/pki -ls | sort -k 11" "katello_pki_dir"
 
@@ -45,12 +45,17 @@ add_files /etc/splice/*
 add_files /etc/httpd/conf.d/splice.conf
 
 # Candlepin
-add_files /var/log/candlepin/*
-add_files /var/log/tomcat6/*
-add_files /var/log/tomcat/*
+add_files /var/log/candlepin/audit*.log*
+add_files /var/log/candlepin/candlepin*.log*
+add_files /var/log/candlepin/cpdb*.log*
+add_files /var/log/candlepin/cpinit*.log*
+add_files /var/log/candlepin/error*.log*
+add_files /var/log/tomcat*/catalina*.log*
+add_files /var/log/tomcat*/host-manager*.log*
+add_files /var/log/tomcat*/localhost*.log*
+add_files /var/log/tomcat*/manager*.log*
 add_files /etc/candlepin/candlepin.conf
-add_files /etc/tomcat6/server.xml
-add_files /etc/tomcat/server.xml
+add_files /etc/tomcat*/server.xml
 
 # Pulp
 add_files /etc/pulp/*.conf
@@ -98,7 +103,10 @@ fi
 # Squid (*)
 if [ $NOGENERIC -eq 0 ]; then
   add_files /etc/squid/squid.conf
-  add_files /var/log/squid/*
+  add_files /var/log/squid/access*.log*
+  add_files /var/log/squid/cache*.log*
+  add_files /var/log/squid/store*.log*
+  add_files /var/log/squid/squid*.out*
 fi
 
 # Qpid Debug


ehelms:
[test]

cfouant:
I tried this out, using your box from boxes.yaml.example and got the following error:

msg: The conditional check 'skip_installer == False' failed. The error was: error while evaluating conditional (skip_installer == False): 'skip_installer' is undefined

The offending line appears to be:
- name: 'Run devel installer'

pmoravec:
@cfouant is your comment relevant to this PR? The change requires https://github.com/theforeman/foreman/pull/3881 applied to `foreman-debug`.

Anyway I will update that PR per @lzap's feedback and consequently this PR as well.

stbenjam:
Hi @pmoravec Can you help me understand what this is for? The redmine ticket and the description of the PR doesn't explain the relevance of the ticket to Katello. The changs don't seem to have to do anything with file sizes?

From what I see here, a huge nack from me to inclusion of specific file names in the directories, if logs get renamed, we'll miss them.  Why is that being done?

Lastly, Satellite references don't belong in upstream. If the commercial product needs something not gathered by katello's debug plugin it should have one of it's own.

pmoravec:
Hi @stbenjam The original aim is/was `foreman-debug` (that is the only/primary way of calling `katello-debug`) to collect newest complete logfiles, instead of current tails of each logs. See original http://projects.theforeman.org/issues/16680 for more.

The change requires different implementation of `add_files` method in `foreman-debug` that `katello-debug` uses. Newly, the method does not collect (tail of) each and every file in arguments, but the method newly sorts the files per modification timestamp and collects up to 10MB of the newest files. That means, if e.g. `/var/log/candlepin/cpinit.log` is 1 month old, newer `/var/log/candlepin/candlepin*.log*` files are bigger than 10MB, just calling

```
add_files /var/log/candlepin/*.log*
```

would skip collecting `cpinit*` logfiles. I wanted to offer a new `foreman-debug` method to apply the size limit per different log type, but that was denied by @lzap in https://github.com/theforeman/foreman/pull/3881#discussion_r82954733 . That's why it is necessary to enumerate different logfiles we need to capture.

Additionally (little bit independently on the original PR), `/var/lob/foreman/dynflow*` was identified as `katello` related logfile, so `katello-debug` shall collect it, instead of `foreman-debug itself.

(reviewing the patch, I think it is safe to merge:

```
add_files /var/log/tomcat6/catalina*.log*
```

with

```
add_files /var/log/tomcat/catalina*.log*
```

to

```
add_files /var/log/tomcat*/catalina*.log*
```

since either `tomcat` xor `tomcat6` logs are present on a system, but not both. Equivalently other logfile types. Let me know if you agree with it and I will update the PR)

stbenjam:
I think this is a huge step back, as now you won't pull potentially relevant log files for free.  foreman-debug by itself won't gather scenario log files from other installer scenarios.  It seems like we could have better logic in foreman-debug that groups files based on a prefix (say 3 characters or so), and ensures to gather a certain amount from each.

Regardless, dynflow doesn't belong here.   There are other users than Katello of foreman-tasks, so there needs to be a foreman-tasks foreman-debug plugin that has this and any other relevant tasks logs.

Likewise, references to Satellite need to be removed from this PR and you'll need to create a foreman-debug plugin for it.

lzap:
> I think this is a huge step back, as now you won't pull potentially relevant log files for free. foreman-debug by itself won't gather scenario log files from other installer scenarios. It seems like we could have better logic in foreman-debug that groups files based on a prefix (say 3 characters or so), and ensures to gather a certain amount from each.

Verbosity is not always a benefit, we actually want to be sure that what we get is what we expect. In this case, changes in foreman-debug allows to define log files with much better granularity. Before the change, some reports were unusable (cut by some limit - regardless if this was implemented as lines count or total file sizes). You will always miss something. With the new approach, you can be always sure you get at least the most relevant (last modified) log file for each `add_files` statement. And this is huge benefit actually!

This may be less comfortable long term, but having usable output is more important.

> Regardless, dynflow doesn't belong here. There are other users than Katello of foreman-tasks, so there needs to be a foreman-tasks foreman-debug plugin that has this and any other relevant tasks logs.

My falut, I told @pmoravec to put it here because I thought it is Katello only. Pull it out into separate plugin please. The same way as Katello does it. Packaging changes are necessary.

> Likewise, references to Satellite need to be removed from this PR and you'll need to create a foreman-debug plugin for it.

I agree this was caused by some changes in the installer recently. I think the best RPM package for this is the satellite branding package. Please make a PR into it. Feel free to ping me if you need guidance.

stbenjam:
> Verbosity is not always a benefit, we actually want to be sure that what we get is what we expect. In this case, changes in foreman-debug allows to define log files with much better granularity. Before the change, some reports were unusable (cut by some limit - regardless if this was implemented as lines count or total file sizes). You will always miss something. With the new approach, you can be always sure you get at least the most relevant (last modified) log file for each add_files statement. And this is huge benefit actually!
> 
> This may be less comfortable long term, but having usable output is more important.

I understand the goal and agree it's useful, I just think it should be implemented differently.  It's very difficult to maintain with this change.  Every package that provides a scenario shouldn't need to build a foreman-debug plugin to include it's log, foreman-debug should grab the installer logs from all scenarios.  When the proxy scenario is merged, someone is going to miss it's log and we'll have to update foreman-debug and wait for a release.  It'll be happening _constantly_ for things like that.

I think there's better ways to solve it, like grabbing everything that matches a glob, checking if the size is over the limit, and then dopping files until it's under the limit, all the while ensuring that at least each 1 file with each X-length prefix name is included.

lzap:
> I understand the goal and agree it's useful, I just think it should be implemented differently. It's very difficult to maintain with this change. Every package that provides a scenario shouldn't need to build a foreman-debug plugin to include it's log, foreman-debug should grab the installer logs from all scenarios. When the proxy scenario is merged, someone is going to miss it's log and we'll have to update foreman-debug and wait for a release. It'll be happening constantly for things like that.
> 
> I think there's better ways to solve it, like grabbing everything that matches a glob, checking if the size is over the limit, and then dopping files until it's under the limit, all the while ensuring that at least each 1 file with each X-length prefix name is included.

Hmm if the installer is our only concern, then I think we are wasting our time here. We can simply use one glob for all installer logs, I doubt they are over 10 MB of size in total, usually there are only several installer runs. E.g. clean 6.2 installation (Foreman with Katello and other plugins) is:

```
 2.4M    /var/log/foreman-installer/
```

on my test instance.

TL;DR - I think we can leave the `add_files /var/log/{katello,capsule}-installer/*` rule as is, @pmoravec

katello/katello-debug.sh
```diff
@@ -45,12 +45,21 @@ add_files /etc/splice/*
...
+add_files /var/log/candlepin/cpinit*.log*
+add_files /var/log/candlepin/error*.log*
+add_files /var/log/tomcat*/catalina*.log*
+add_files /var/log/tomcat*/host-manager*.log*
+add_files /var/log/tomcat*/localhost*.log*
+add_files /var/log/tomcat*/manager*.log*
+add_files /var/log/tomcat/catalina*.log*
+add_files /var/log/tomcat/host-manager*.log*
+add_files /var/log/tomcat/localhost*.log*
+add_files /var/log/tomcat/manager*.log*
```
stbenjam: These duplicate the versions above, e.g. `/var/log/tomcat*/catalina*.log*`  matches `/var/log/tomcat/catalina*.log`

stbenjam:
Well it doesn't really address my concern about foreman, since when there's eventually a proxy scenario someone's going forget to add it to debug.  This is _much_ harder to maintain now.

From the katello perspective this PR is fine I guess, just one comment about duplicate tomcat globbing matches.

katello/katello-debug.sh
pmoravec: Thanks for spotting it, my fault, now fixed.

stbenjam:
Thanks [test]

lzap:
Core counterpart was merged, @stbenjam thanks for review. If there is a glob that adds all scenarios I think it would be easy to add it. It could be on one line, users usually install only one scenario, that will fit into 10 MB limit

```
add_files /var/log/*-installer/*
```

pmoravec:
Updated the PR by calling add_files to `*-installer` on 2 places (`/etc` and logs)

stbenjam:
The *-installer isn't really needed, everything goes to foreman-installer, just the file names in the directories are different for the scenario. That's why I was complaining about the move in foreman-debug to only include the foreman scenario files...

Anyway, please update this to capture

```
/var/log/foreman-installer/katello*log
/var/log/foreman-installer/capsule*log
```

If you want you can include *log but that kind of misses my main point, it's not about one set of files, but whatever, the foreman side merged, just fix this and I'll merge it too.

lzap:
As per IRC discussion with @stbenjam it looks like "everything goes to /etc/foreamn-installer and /var/log/foreman-installer", sorry for the noise. The "*-installer" is unnecessary I assume.

pmoravec:
. @stbenjam is something further pending / worth improving in the PR? (and can you please kick off automated tests if not)?

stbenjam:
Oops, sorry, looks good to go.

[test]

stbenjam:
Thanks @pmoravec !

PR #298, opened 2016-09-28
fixes #16732 - gzips archive after services turned back on

cfouant:


stbenjam:
Looks good to me, just going to test it real quick.

stbenjam:
Works great, ACK

PR #291, opened 2016-09-19
Fixes #16586 - restart gofer with katello-service

chris1984:

PR #306, opened 2016-10-12
Fixes #16251 - Add katello-service disable/enable options

sean797:
I've also added goferd to the bash_completion script, it was missed from #291 :-)

### DIFF
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -2,7 +2,7 @@
 
 require 'optparse'
 
-ACTIONS = ['restart', 'stop', 'start', 'status', 'list']
+ACTIONS = ['restart', 'stop', 'start', 'status', 'list', 'enable', 'disable']
 COMMAND = '/usr/sbin/service-wait'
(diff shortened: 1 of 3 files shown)


katello/katello-service
```diff
@@ -50,7 +50,7 @@ end
 
 def service_exists?(service)
   upstart = File.exist?("/etc/init.d/#{service}")
-  systemd = `systemctl is-enabled #{service} 2>&1` && $?.success?
+  systemd = `systemctl cat #{service} 2>&1` && $?.success?
```
ehelms, line 53: Curiosity -- why switch from is-enabled to cat?

sean797: If a service is disabled `systemctl is-enabled #{service}` will exit with 1. So enabling a disabled service would fail silently as it would think the service doesn't exist. `systemctl cat` seems like the best way to test if a service exists.

ehelms:
ACK - thanks @sean797 !

PR #307, opened 2016-10-14
Fixes #16927 - Dir.exist? isnt available on el6 (ruby 1.8)

sean797:


ehelms:
ACK - thanks @sean797 !

PR #311, opened 2016-10-20
Fixes #17047 - Redirect cron to /dev/null

chris1984:


stbenjam:
Looks good to me, APJ

PR #312, opened 2016-10-20
Cherry picks 3.2 rc3

johnpmitsch:


ehelms:
The failure is curious -- http://koji.katello.org/koji/getfile?taskID=551576&name=root.log -- @jlsherrill recognize this? 

We should be able to generally ignore it given katello-agent has already been built for 2.7.0 and will just get tagged over.

jlsherrill:
Yeah, I think I got this when I was building at some point.  I think a
Regen repo fixed it, but I would just remove this commit from the pr and
cherry pick as you mentioned

jlsherrill:
actually the comps changes from the katello-agent commit is required.  

I'm fine with merging as is.

PR #315, opened 2016-11-03
refs #17201 - alter sed so it works w/ any YAML hash style

stbenjam:
It's currently fixed to matching ['x', 'y'].  This changes to sed the
paths directly instead.

Because of https://github.com/Katello/katello-installer/pull/411


stbenjam:
Hm, does this really need a different ticket? The two changes go together.

ehelms:
@stbenjam nah this is fine

stbenjam:
Take that @theforeman-bot :stuck_out_tongue_closed_eyes:

stbenjam:
@ehelms Can you review again? BATS is broken until the sed gets fixed: http://ci.theforeman.org/view/Katello%20Pipeline/job/release_test_katello_nightly_el7/68/

ehelms:
ACK

PR #316, opened 2016-11-07
Fixes #17254 - remove katello-common last

sean797:


stbenjam:
I'd prefer if we didn't say "Success!" and then do another thing.  Success should indicate whether everything succeed or failed.

sean797:
@stbenjam that makes sense! Updated :+1:

stbenjam:
So if I understand correctly, this is supposed to prevent removing katello-remove if something failed? 

I don't see any checks if the other commands failed, so you'll still reach the end  and remove katello-common regardless of what happens.

sean797:
Ah sorry @stbenjam that was kind of stupid of me. Updated again :-)

ehelms:
@sean797 sorry for the delay, this will need a rebase

sean797:
Closing, I don't think the original error will happen now that it's been re-written in Ruby.

PR #318, opened 2016-11-16
Refs #17230 - added tracer support

sean797:
Still waiting for `tracer-0.6.12` to be released. We can check the status from https://bodhi.fedoraproject.org/updates/?packages=tracer


jlsherrill:
hah, didn't even realize this was open (but hadn't checked).  I've bumped the tag on katello-agent, so it may build now.

jlsherrill:
[test]

katello-agent/katello-agent.spec
```diff
@@ -33,6 +33,11 @@ Requires: yum-plugin-security
 Requires: yum-security
 %endif
 
+%if 0%{?fedora} > 18 || 0%{?rhel} > 6
+Requires: python2-tracer >= 1.6.12 
+Requires: python-rhsm
+%endif
```
jlsherrill, line 39: curious if you had thought about making this a subpackage that is only built on those platforms?

jlsherrill:
ah right, you'll need to download https://github.com/Katello/katello-agent/archive/2.8.0.tar.gz and add it via git annex add 2.8.0  (named as 2.8.0 due to the silly naming that github gives us via redirection)

katello-agent/katello-agent.spec
sean797: Create a katello-agent-tracer pkg? I guess we could do but no I hadn't thought about it.

jlsherrill:
ACK thanks @sean797 !

PR #323, opened 2016-11-29
Fixes #17522 - Add katello-change-hostname command

johnpmitsch:
This adds a command to go through the necessary steps to change the hostname in katello.

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -2,11 +2,12 @@
 
 require "socket"
 require "optparse"
-require "highline/import"
+require 'yaml'
 
-DEFAULT_PROGRAM = "foreman"
-DEFAULT_SCENARIO = "katello"
-DEFAULT_SSL_DIR = "/root/ssl-build"
+raise 'Must run as root' unless Process.uid == 0
+
+DEFAULT_SCENARIO = File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
+DEFAULT_SCENARIO == "katello" ? DEFAULT_PROGRAM = "foreman" : DEFAULT_PROGRAM = DEFAULT_SCENARIO
 
 def is_rhel_version?(version)
   `cat /etc/redhat-release`[/(\d+\.)/].include?(version)
@@ -25,16 +26,33 @@ def get_hostname
   hostname.chomp 
 end
 
-response = agree("This will modify your system. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]")
-unless response
-  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
-  exit
+def has_disable_system_check_option?
+  katello_installer_rpm = `rpm -qa | grep katello-installer`
+  katello_installer_version = katello_installer_rpm[/(\d+\.)(\d+\.)(\d+)/]
+  Gem::Version.new(katello_installer_version) >= Gem::Version.new("3.2.0")
+end
+
+def yesno
+  begin
+    system("stty raw -echo")
+    str = STDIN.getc
+  ensure
+    system("stty -raw echo")
+  end
+  if str.chr.downcase == "y"
+    return true
+  elsif str.chr.downcase == "n"
+    return false
+  else
+    puts "Invalid Character. Try again: [y/n]"
+    yesno
+  end
 end
 
 options = {}
 options[:program] = DEFAULT_PROGRAM
 options[:scenario] = DEFAULT_SCENARIO
-options[:ssl_dir] = DEFAULT_SSL_DIR
+options[:system_check] = false
 
 opt_parser = OptionParser.new do |opt|
   opt.banner = "Usage: katello-change-hostname HOSTNAME [OPTIONS]"
@@ -44,11 +62,11 @@ opt_parser = OptionParser.new do |opt|
   opt.separator  ""
   opt.separator  "Options"
 
-  opt.on("-u","--username USERNAME","admin username REQUIRED") do |username|
+  opt.on("-u","--username USERNAME","admin username (required)") do |username|
     options[:username] = username
   end
 
-  opt.on("-p","--password PASSWORD","admin password REQUIRED") do |password|
+  opt.on("-p","--password PASSWORD","admin password (required)") do |password|
     options[:password] = password
   end
 
@@ -56,16 +74,15 @@ opt_parser = OptionParser.new do |opt|
     options[:program] = program
   end
 
-  opt.on("-s","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
+  opt.on("-S","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
     options[:scenario] = scenario
   end
 
-  opt.on("-d","--disable-system-checks","runs the installer with --disable-system-checks") do |system_check|
-    options[:system_check] = true
-  end
 
-  opt.on("-l","--ssl-dir SSL_DIR","specify the ssl cert directory to replace (defaults to #{DEFAULT_SSL_DIR})") do |ssl_dir|
-    options[:ssl_dir] = ssl_dir
+  if has_disable_system_check_option?
+    opt.on("-d","--disable-system-checks","runs the installer with --disable-system-checks") do |system_check|
+      options[:system_check] = true
+    end
   end
 
   opt.on("-h","--help","help") do
@@ -75,7 +92,12 @@ opt_parser = OptionParser.new do |opt|
 end
 opt_parser.parse!
 
-raise 'Must run as root' unless Process.uid == 0
+STDOUT.print("This will modify your system. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]:")
+response = yesno
+unless response
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
 
 unless options[:username] && options[:password]
   STDOUT.puts "Username and/or Password options are missing!"
@@ -83,13 +105,9 @@ unless options[:username] && options[:password]
   exit
 end
 
-if options[:scenario] == "katello"
-  proxy = "Smart Proxy"
-  plural_proxy = "Smart Proxies"
-else
-  proxy = "Capsule"
-  plural_proxy = "Capsules"
-end
+# kept as variables for easy changing in backports
+proxy = "Foreman Proxy"
+plural_proxy = "Foreman Proxies"
 
 if ARGV[0] && ARGV.count >= 1
   new_hostname = ARGV[0]
@@ -101,9 +119,14 @@ end
 # Get the hostname from your system
 old_hostname = get_hostname
 
+scenario_answers = YAML.load_file("/etc/foreman-installer/scenarios.d/#{options[:scenario]}-answers.yaml")
+ssl_build_dir = scenario_answers["certs"]["ssl_build_dir"]
+ssl_backup_dir = "#{ssl_build_dir}-#{timestamp}-#{old_hostname}"
+
 STDOUT.puts "Updating default #{proxy}"
 default_capsule_id = `hammer -u #{options[:username]} -p #{options[:password]} capsule list | grep #{old_hostname} | awk '{ print $1 }'`[0]
-`hammer -u #{options[:username]} -p #{options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname} >> /dev/null`
+# Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
+`hammer -u #{options[:username]} -p #{options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname} 2> /dev/null`
 
 STDOUT.puts "Updating installation media paths"
 installation_media_ids = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $1 }'`.split("\n")
@@ -139,20 +162,27 @@ STDOUT.puts "stopping services"
 
 STDOUT.puts "deleting old certs"
 `
-rm -rf /etc/pki/katello{,.bak}
+rm -rf #{scenario_answers["certs"]["pki_dir"]}{,.bak}
 rm -rf /etc/pki/katello-certs-tools{,.bak}
 rm -rf /etc/candlepin/certs/amqp{,.bak}
-rm -rf /etc/foreman-proxy/*ssl*
+rm -rf #{scenario_answers["foreman_proxy"]["ssl_ca"]}
+rm -rf #{scenario_answers["foreman_proxy"]["ssl_cert"]}
+rm -rf #{scenario_answers["foreman_proxy"]["ssl_key"]}
+rm -rf #{scenario_answers["foreman_proxy"]["foreman_ssl_ca"]}
+rm -rf #{scenario_answers["foreman_proxy"]["foreman_ssl_cert"]}
+rm -rf #{scenario_answers["foreman_proxy"]["foreman_ssl_key"]}
 rm -rf /etc/foreman/old-certs
-rm -rf /etc/foreman/*.pem
-rm -rf /var/lib/puppet/ssl
+rm -rf #{scenario_answers["foreman"]["client_ssl_ca"]}
+rm -rf #{scenario_answers["foreman"]["client_ssl_cert"]}
+rm -rf #{scenario_answers["foreman"]["client_ssl_key"]}
+rm -rf #{scenario_answers["foreman_proxy"]["ssldir"]}
 rm -rf /etc/pki/katello/nssdb
 rm -f /etc/tomcat/keystore
 rm -f /etc/pki/katello/keystore
-cp -r #{options[:ssl_dir]} #{options[:ssl_dir]}.#{old_hostname}#{timestamp}
-rm -rf #{options[:ssl_dir]}
+cp -r #{ssl_build_dir} #{ssl_backup_dir}
+rm -rf #{ssl_build_dir}
 `
-STDOUT.puts "backed up #{options[:ssl_dir]} to #{options[:ssl_dir]}.#{old_hostname}#{timestamp}"
+STDOUT.puts "backed up #{ssl_build_dir} to #{ssl_backup_dir}"
 STDOUT.puts "updating hostname in /etc/hosts"
 `sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hosts`
 
@@ -170,17 +200,24 @@ installer = "#{options[:program]}-installer --scenario #{options[:scenario]} -v 
 installer << " --disable-system-checks" if options[:system_check]
 STDOUT.puts installer
 installer_output = `#{installer}`
+installer_success = $?.success?
 STDOUT.puts installer_output
 
-STDOUT.puts "
-Hostname change complete!
+if installer_success
+  STDOUT.puts "
+  Hostname change complete!
+  
+  You will need to re-register any #{plural_proxy} or clients with the server. 
+  
+  When re-registering clients, you will have to reinstall the bootstrap RPM.
+  Any mention of the #{old_hostname} on your clients will have to change to #{new_hostname} (i.e. /etc/hosts)
+  
+  For #{plural_proxy}, you will need to regenerate the certs on the #{options[:scenario].capitalize} server and reinstall the #{proxy}
+  
+  If you want to use custom certificates, re-run the #{options[:program]}-installer with custom certificate options
 
-You will need to re-register any #{plural_proxy} or clients with the #{options[:scenario].capitalize} server. 
-
-When re-registering clients, you will have to reinstall the bootstrap RPM.
-Any mention of the #{old_hostname} on your clients will have to change to #{new_hostname} (i.e. /etc/hosts)
-
-For #{plural_proxy}, you will need to regenerate the certs on the #{options[:scenario].capitalize} server and reinstall the #{proxy}
-
-Short hostnames have not been updated, please update those manually.
-"
+  Short hostnames have not been updated, please update those manually.
+  "
+else
+  STDOUT.puts "Something went wrong with the #{options[:scenario].capitalize} installer! Please check the above output and the corresponding logs"
+end
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -93,7 +93,11 @@ opt_parser = OptionParser.new do |opt|
 end
 opt_parser.parse!
 
-STDOUT.print("This will modify your system. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]:")
+# kept as variables for easy changing in backports
+proxy = "Foreman Proxy"
+plural_proxy = "Foreman Proxies"
+
+STDOUT.print("This will modify your system. You will need to re-register any #{ plural_proxy } and #{options[:scenario].capitalize} clients after script completion. #{ plural_proxy } will have to be reinstalled. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and want to proceed with changing your hostname? [y/n]:")
 response = yesno
 unless response
   STDOUT.puts "Hostname change aborted, no changes have been made to your system"
@@ -105,10 +109,6 @@ unless options[:username] && options[:password]
   puts opt_parser
   exit
 end
-
-# kept as variables for easy changing in backports
-proxy = "Foreman Proxy"
-plural_proxy = "Foreman Proxies"
 
 if ARGV[0] && ARGV.count >= 1
   new_hostname = ARGV[0]
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -1,0 +1,186 @@
+#!/usr/bin/ruby
+
+require "socket"
+require "optparse"
+require "highline/import"
+
+DEFAULT_PROGRAM = "foreman"
+DEFAULT_SCENARIO = "katello"
+DEFAULT_SSL_DIR = "/root/ssl-build"
+
+def is_rhel_version?(version)
+  `cat /etc/redhat-release`[/(\d+\.)/].include?(version)
+end
+
+def timestamp
+  Time.now.strftime("%Y%m%d%H%M")
+end
+
+def get_hostname
+  if is_rhel_version?("6")
+    hostname = `hostname`
+  else
+    hostname = Socket.gethostname
+  end
+  hostname.chomp 
+end
+
+response = agree("This will modify your system. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]")
+unless response
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
+
+options = {}
+options[:program] = DEFAULT_PROGRAM
+options[:scenario] = DEFAULT_SCENARIO
+options[:ssl_dir] = DEFAULT_SSL_DIR
+
+opt_parser = OptionParser.new do |opt|
+  opt.banner = "Usage: katello-change-hostname HOSTNAME [OPTIONS]"
+  opt.separator  ""
+  opt.separator  "Example:"
+  opt.separator  " katello-change-hostname foo.example.com -u admin -p changeme"
+  opt.separator  ""
+  opt.separator  "Options"
+
+  opt.on("-u","--username USERNAME","admin username REQUIRED") do |username|
+    options[:username] = username
+  end
+
+  opt.on("-p","--password PASSWORD","admin password REQUIRED") do |password|
+    options[:password] = password
+  end
+
+  opt.on("-g","--program PROGRAM","name of the program you are modifying (defaults to #{DEFAULT_PROGRAM})") do |program|
+    options[:program] = program
+  end
+
+  opt.on("-s","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
+    options[:scenario] = scenario
+  end
+
+  opt.on("-d","--disable-system-checks","runs the installer with --disable-system-checks") do |system_check|
+    options[:system_check] = true
+  end
+
+  opt.on("-l","--ssl-dir SSL_DIR","specify the ssl cert directory to replace (defaults to #{DEFAULT_SSL_DIR})") do |ssl_dir|
+    options[:ssl_dir] = ssl_dir
+  end
+
+  opt.on("-h","--help","help") do
+    puts opt_parser
+    exit
+  end
+end
+opt_parser.parse!
+
+raise 'Must run as root' unless Process.uid == 0
+
+unless options[:username] && options[:password]
+  STDOUT.puts "Username and/or Password options are missing!"
+  puts opt_parser
+  exit
+end
+
+if options[:scenario] == "katello"
+  proxy = "Smart Proxy"
+  plural_proxy = "Smart Proxies"
+else
+  proxy = "Capsule"
+  plural_proxy = "Capsules"
+end
+
+if ARGV[0] && ARGV.count >= 1
+  new_hostname = ARGV[0]
+else
+  puts opt_parser
+  exit
+end
+
+# Get the hostname from your system
+old_hostname = get_hostname
+
+STDOUT.puts "Updating default #{proxy}"
+default_capsule_id = `hammer -u #{options[:username]} -p #{options[:password]} capsule list | grep #{old_hostname} | awk '{ print $1 }'`[0]
+`hammer -u #{options[:username]} -p #{options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname} >> /dev/null`
+
+STDOUT.puts "Updating installation media paths"
+installation_media_ids = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $1 }'`.split("\n")
+installation_media_paths = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $5 }'`.split("\n")
+installation_media_ids.zip(installation_media_paths).each do |id, path|
+  `hammer -u #{options[:username]} -p #{options[:password]} medium update --id #{id} --path #{path.sub! old_hostname, new_hostname}`   
+end
+
+if is_rhel_version?("6")
+  STDOUT.puts "updating hostname in /etc/sysconfig/network"
+  STDOUT.puts("sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/sysconfig/network")
+  `sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/sysconfig/network`
+  STDOUT.puts "setting hostname"
+  `hostname #{new_hostname}`
+elsif is_rhel_version?("7")
+  STDOUT.puts "updating hostname in /etc/hostname"
+  `sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hostname`
+  STDOUT.puts "setting hostname"
+  `hostnamectl set-hostname #{new_hostname}`
+else
+  STDOUT.puts "couldn't find RHEL version: #{RHEL_VERSION} - exiting, no changes have been made to your system"
+  exit
+end
+
+STDOUT.puts "checking if hostname was changed"
+if get_hostname != new_hostname
+  STDOUT.puts "the new hostname was not changed successfully, exiting script"
+  exit
+end
+
+STDOUT.puts "stopping services"
+`katello-service stop`
+
+STDOUT.puts "deleting old certs"
+`
+rm -rf /etc/pki/katello{,.bak}
+rm -rf /etc/pki/katello-certs-tools{,.bak}
+rm -rf /etc/candlepin/certs/amqp{,.bak}
+rm -rf /etc/foreman-proxy/*ssl*
+rm -rf /etc/foreman/old-certs
+rm -rf /etc/foreman/*.pem
+rm -rf /var/lib/puppet/ssl
+rm -rf /etc/pki/katello/nssdb
+rm -f /etc/tomcat/keystore
+rm -f /etc/pki/katello/keystore
+cp -r #{options[:ssl_dir]} #{options[:ssl_dir]}.#{old_hostname}#{timestamp}
+rm -rf #{options[:ssl_dir]}
+`
+STDOUT.puts "backed up #{options[:ssl_dir]} to #{options[:ssl_dir]}.#{old_hostname}#{timestamp}"
+STDOUT.puts "updating hostname in /etc/hosts"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hosts`
+
+STDOUT.puts "updating hostname in foreman installer scenarios"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml`
+
+STDOUT.puts "removing last_scenario.yml file"
+`rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml`
+
+STDOUT.puts "re-running the installer"
+
+installer = "#{options[:program]}-installer --scenario #{options[:scenario]} -v --certs-regenerate-ca=true --certs-regenerate=true \
+             --foreman-proxy-register-in-foreman true"
+
+installer << " --disable-system-checks" if options[:system_check]
+STDOUT.puts installer
+installer_output = `#{installer}`
+STDOUT.puts installer_output
+
+STDOUT.puts "
+Hostname change complete!
+
+You will need to re-register any #{plural_proxy} or clients with the #{options[:scenario].capitalize} server. 
+
+When re-registering clients, you will have to reinstall the bootstrap RPM.
+Any mention of the #{old_hostname} on your clients will have to change to #{new_hostname} (i.e. /etc/hosts)
+
+For #{plural_proxy}, you will need to regenerate the certs on the #{options[:scenario].capitalize} server and reinstall the #{proxy}
+
+Short hostnames have not been updated, please update those manually.
+"
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -2,6 +2,7 @@
 
 require "socket"
 require "optparse"
+require "rubygems"
 require 'yaml'
 
 raise 'Must run as root' unless Process.uid == 0
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -8,7 +8,7 @@ require 'yaml'
 raise 'Must run as root' unless Process.uid == 0
 
 DEFAULT_SCENARIO = File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
-DEFAULT_SCENARIO == "katello" ? DEFAULT_PROGRAM = "foreman" : DEFAULT_PROGRAM = DEFAULT_SCENARIO
+DEFAULT_PROGRAM = DEFAULT_SCENARIO == "katello" ? "foreman" : DEFAULT_SCENARIO
 
 def is_rhel_version?(version)
   `cat /etc/redhat-release`[/(\d+\.)/].include?(version)
@@ -97,7 +97,12 @@ opt_parser.parse!
 proxy = "Foreman Proxy"
 plural_proxy = "Foreman Proxies"
 
-STDOUT.print("This will modify your system. You will need to re-register any #{ plural_proxy } and #{options[:scenario].capitalize} clients after script completion. #{ plural_proxy } will have to be reinstalled. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and want to proceed with changing your hostname? [y/n]:")
+STDOUT.print("This will modify your system. You will need to re-register any #{ plural_proxy } and " \
+             "#{options[:scenario].capitalize} clients after script completion. #{ plural_proxy } " \
+             "will have to be reinstalled. If you are using custom certificates, you will have to run the " \
+             "#{options[:program]}-installer again with custom certificate options after this script completes." \
+             " Have you taken the necessary precautions (backups, snapshots, etc...) and want to proceed with " \
+             "changing your hostname? [y/n]:")
 response = yesno
 unless response
   STDOUT.puts "Hostname change aborted, no changes have been made to your system"
--- a/katello/katello.spec
+++ b/katello/katello.spec
@@ -24,6 +24,7 @@ Source7:    katello-backup
 Source8:    katello-service-bash_completion.sh
 Source9:    qpid-core-dump
 Source10:   katello-clean-empty-puppet-environments 
+Source11:   katello-change-hostname
 
 BuildRequires: asciidoc
 BuildRequires: util-linux
@@ -85,6 +86,7 @@ install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.weekly/katello-clean-
 # install important scripts
 mkdir -p %{buildroot}%{_bindir}
 mkdir -p %{buildroot}%{_sbindir}
+install -Dp -m0755 %{SOURCE11} %{buildroot}%{_bindir}/katello-change-hostname
 install -Dp -m0755 %{SOURCE9} %{buildroot}%{_bindir}/qpid-core-dump
 install -Dp -m0755 %{SOURCE7} %{buildroot}%{_bindir}/katello-backup
 install -Dp -m0755 %{SOURCE6} %{buildroot}%{_bindir}/katello-restore
@@ -127,6 +129,7 @@ Common runtime components of %{name}
 %{_bindir}/katello-backup
 %{_bindir}/katello-restore
 %{_bindir}/qpid-core-dump
+%{_bindir}/katello-change-hostname
 %config(missingok) %{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
 %config(missingok) %{_sysconfdir}/cron.weekly/katello-remove-orphans
 


katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+`
+
+STDOUT.puts "Restarting services"
+`katello-service restart`
+
+STDOUT.puts "
+Hostname change complete!
+
+If you have a capsule you will need to reregister it with RHSM and then run the following (replacing <capsule-hostname> with your capsule's hostname):
+capsule-certs-generate --capsule-fqdn '<capsule-hostname>' --certs-tar '~/<capsule-hostname>-certs.tar'
```
jlsherrill: Is re-registration actually required?  Or just updating the bootstrap script?

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+STDOUT.puts "Starting the hostname change..."
+
+STDOUT.puts("\nA hostname change requires a manual change in Satellite. \n
+            Please log into UI and rename the default capsule (under infrastructure->capsules) 
+            to the new hostname using both the name and url fields.")
+
+unless HighLine.agree('Have you made this necessary change and want to proceed with the hostname change? [yes/no]')
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
```
jlsherrill, line 110: since it appears that the changes below cause a new CA to be generated, should we also warn that all clients will need their bootstrap rpm upgraded/re-installed?

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+
+if ARGV[0] && ARGV.count == 1
+  NEW_HOSTNAME = ARGV[0]
+else
+  STDOUT.puts "USAGE: katello-change-hostname <new-hostname>"
+  exit
+end
+
+# Get the hostname from your system
+OLD_HOSTNAME = Socket.gethostname
```
jlsherrill: should we check here that OLD_HOSTNAME != the new hostname?  In case the user thought they should go ahead and do that manually (which would cause future actions to fail).

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+
+STDOUT.puts "checking if hostname was changed"
+hostname = Socket.gethostname
+if hostname != NEW_HOSTNAME
+  STDOUT.puts "the new hostname was not changed, check hostname -f"
+  exit
+end
+
+STDOUT.puts "updating hostname in /etc/foreman-installer/scenarios.d/satellite-answers.yaml"
+`sed -i -e 's/#{OLD_HOSTNAME}/#{NEW_HOSTNAME}/g' /etc/foreman-installer/scenarios.d/satellite-answers.yaml`
```
jlsherrill: out of curiosity can this not be accomplished via installer options?

katello/katello-change-hostname
johnpmitsch: +1

katello/katello-change-hostname
johnpmitsch: yes, we should mention that, @chris1984 was trying to figure out a way to not change the CA, maybe we can do that instead

katello/katello-change-hostname
johnpmitsch: Unsure, @ehelms or @stbenjam  do you know?

johnpmitsch:
To run on rhel6 production katello machine:

`scl enable tfm 'ruby /path/to/change_hostname.rb foo.example.com'`

katello/katello-change-hostname
ehelms: You could use Facter instead of Socket

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+  STDOUT.puts "USAGE: katello-change-hostname <new-hostname>"
+  exit
+end
+
+# Get the hostname from your system
+OLD_HOSTNAME = Socket.gethostname
+
+RHEL_VERSION = `rpm -q --queryformat '%{VERSION}' redhat-release-server`[0]
+
+STDOUT.puts "stopping services"
```
ehelms: We might should warn users earlier in the process of this script that it will bring down all services for a maintenance period.

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/ruby
+
+require "socket"
+require 'highline/import'
+
+STDOUT.puts "Starting the hostname change..."
+
+STDOUT.puts("\nA hostname change requires a manual change in Satellite. \n
```
ehelms: s/Satellite/Foreman

ehelms: Does a user do this rename before they run this script or after?

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+
+# Get the hostname from your system
+OLD_HOSTNAME = Socket.gethostname
+
+RHEL_VERSION = `rpm -q --queryformat '%{VERSION}' redhat-release-server`[0]
+
+STDOUT.puts "stopping services"
+`katello-service stop`
+
+STDOUT.puts "deleting old certs"
```
ehelms: Should this backup the `/root/ssl-build` directory for users or should we warn them to do so before they start?

katello/katello-change-hostname
ehelms: I would think it could, you would have to know each option that needs changing which might be burdensome. However, a hook in the installer would be able to load the answers in and make this change I'd think.

ehelms: Also this all about some `satellite-answers` which does not exist.

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+hostname = Socket.gethostname
+if hostname != NEW_HOSTNAME
+  STDOUT.puts "the new hostname was not changed, check hostname -f"
+  exit
+end
+
+STDOUT.puts "updating hostname in /etc/foreman-installer/scenarios.d/satellite-answers.yaml"
+`sed -i -e 's/#{OLD_HOSTNAME}/#{NEW_HOSTNAME}/g' /etc/foreman-installer/scenarios.d/satellite-answers.yaml`
+
+STDOUT.puts "re-running the Satellite installer"
```
ehelms: s/Satellite/Foreman

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+`
+satellite-installer --scenario satellite -v --certs-regenerate-ca=true --certs-regenerate=true \
+  --foreman-foreman-url "https://#{NEW_HOSTNAME}" \
+  --foreman-servername #{NEW_HOSTNAME} \
+  --capsule-qpid-router-broker-addr  #{NEW_HOSTNAME}\
+  --certs-ca-common-name #{NEW_HOSTNAME} \
+  --certs-node-fqdn #{NEW_HOSTNAME} \
+  --capsule-parent-fqdn #{NEW_HOSTNAME}\
+  --foreman-proxy-register-in-foreman true
```
ehelms: If we already altered the answers file with the new hostname, why do we need to call these options with it?

katello/katello-change-hostname
```diff
@@ -0,0 +1,93 @@
...
+  STDOUT.puts "the new hostname was not changed, check hostname -f"
+  exit
+end
+
+STDOUT.puts "updating hostname in /etc/foreman-installer/scenarios.d/satellite-answers.yaml"
+`sed -i -e 's/#{OLD_HOSTNAME}/#{NEW_HOSTNAME}/g' /etc/foreman-installer/scenarios.d/satellite-answers.yaml`
+
+STDOUT.puts "re-running the Satellite installer"
+`
+satellite-installer --scenario satellite -v --certs-regenerate-ca=true --certs-regenerate=true \
```
ehelms: s/satellite-installer/foreman-installer

ehelms: s/satellite/katello

katello/katello-change-hostname
ehelms: Does this mean the default capsule or ALL external capsules will need to regenerate certificate bundles?

johnpmitsch: @ehelms currently, all external capsules will have to re-register

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/ruby
+# This script changed the hostname of a Satellite 6.2 machine
```
bbuckingham: s/changed/changes/

Also, should references to 'Satellite' (here and below) be 'Katello'?

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
...
+
+unless response
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
+
+if ARGV[0] && ARGV.count == 1
+  new_hostname = ARGV[0]
+else
+  STDOUT.puts "USAGE: ./change_hostname.rb <new-hostname>"
```
bbuckingham: should this reference 'katello-change-hostname' ?

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
...
+  `sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/sysconfig/network`
+  STDOUT.puts "setting hostname"
+  `hostname #{new_hostname}`
+elsif is_rhel_version?("7")
+  STDOUT.puts "updating hostname in /etc/hostname"
+  `sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hostname`
+  STDOUT.puts "setting hostname"
+  `hostnamectl set-hostname #{new_hostname}`
+else
+  STDOUT.puts "couldn't find RHEL version: #{RHEL_VERSION}"
```
bbuckingham: should we do the release checking earlier?  concern would be that we hit this error condition, but now we have altered the user's server.

(note: not likely to happen, since these are the only supported releases for install)

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
...
+  STDOUT.puts "the new hostname was not changed successfully, exiting script"
+  exit
+end
+
+STDOUT.puts "updating hostname in /etc/foreman-installer/scenarios.d/satellite-answers.yaml"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/foreman-installer/scenarios.d/satellite-answers.yaml`
+
+STDOUT.puts "re-running the Satellite installer"
+`
+satellite-installer --scenario satellite -v --certs-regenerate-ca=true --certs-regenerate=true \
```
bbuckingham: foreman-installer --scenario katello ?

bbuckingham:
Would it be useful to have a script that can also change the hostname of a capsule?

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
...
+STDOUT.puts "updating hostname in /etc/hosts"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hosts`
+
+STDOUT.puts "checking if hostname was changed"
+if get_hostname != new_hostname
+  STDOUT.puts "the new hostname was not changed successfully, exiting script"
+  exit
+end
+
+STDOUT.puts "updating hostname in /etc/foreman-installer/scenarios.d/satellite-answers.yaml"
```
ehelms: You could use some of the installer metadata to figure this out probably by using the `/etc/foreman-installer/scenarios.d/last_scenario.yaml` to tell you what the current scenario config is. That does not solve the telling you which installer it is question however.

katello/katello-change-hostname
sean797: It should also get this from the `<scenario>.answers.yaml` file `ssl_build_dir` parameter

katello/katello-change-hostname
```diff
@@ -0,0 +1,129 @@
...
+STDOUT.puts "deleting old certs"
+`
+rm -rfv /etc/pki/katello{,.bak}
+rm -rfv /etc/pki/katello-certs-tools{,.bak}
+rm -rfv /etc/candlepin/certs/amqp{,.bak}
+rm -rfv /etc/foreman-proxy/*ssl*
+rm -rfv /etc/foreman/old-certs
+rm -rfv /etc/foreman/*.pem
+rm -rfv /var/lib/puppet/ssl
+rm -rfv /root/ssl-build
```
sean797: In fact most of these are installer options, it should probably get them from the relevant one in the answer file

johnpmitsch:
To test: Run with --help to display various options and usage. You will have to run it as a script and it doesn't work on dev environments  due to the use of `katello-service` so I recommend testing on 3.2 and 3.1.

Its noted in the script, but you will have to re-register any clients or capsules after the script runs

johnpmitsch:
@bbuckingham @ehelms @jlsherrill @chris1984 @sthirugn - updated, please review and test when you have a chance :)

pgagne:
@johnpmitsch like @bbuckingham said, i think this script should also change the name of the default capsule.

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+  exit
+end
+
+STDOUT.puts("\nA hostname change requires a manual change in the program. " \
+	     "Please log into UI and rename the default #{proxy} (under Infrastructure -> #{proxy}) " \
+             "to #{new_hostname} using both the name and url fields. " \
+             "After this script completes, you will have to re-register any capsules and #{options[:scenario]} clients. " \
+             "All services will be stopped and started during this process. This script will modify your system!")
+STDOUT.print("\nHave you made this necessary change and want to proceed with changing your hostname to #{new_hostname}? [y/n]")
+response = yesno
```
evgeni: why not `agree` or `ask` from `highline/import`?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+installer_output = `#{installer}`
+STDOUT.puts installer_output
+
+STDOUT.puts "Restarting services"
+`katello-service restart`
+
+STDOUT.puts "
+Hostname change complete!
+
+You will need to reregister any clients or capsules with the #{options[:scenarion]} server
```
jlsherrill: is :sceanrion a typo?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+opt_parser = OptionParser.new do |opt|
+  opt.banner = "Usage: katello-change-hostname HOSTNAME [OPTIONS]"
+  opt.separator  ""
+  opt.separator  "Example:"
+  opt.separator  " katello-change-hostname foo.example.com"
+  opt.separator  ""
+  opt.separator  "Options"
+
+  opt.on("-s","--program PROGRAM","name of the scenario you are modifying (defaults to #{DEFAULT_PROGRAM})") do |program|
```
ehelms: This help text seems to be a copy of the scenario help text.

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+  opt.on("-d","--disable-system-checks","runs the installer with --disable-system-checks") do |system_check|
+    options[:system_check] = true
+  end
+
+  opt.on("-l","--ssl-dir SSL_DIR","specify the ssl cert directory to replace (defaults to #{DEFAULT_SSL_DIR})") do |ssl_dir|
+    options[:ssl_dir] = ssl_dir
+  end
+
+ opt.on("-h","--help","help") do
```
ehelms: Indentation is off here.

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+end
+opt_parser.parse!
+
+raise 'Must run as root' unless Process.uid == 0
+
+if options[:scenario] == "katello"
+  proxy = "Smart Proxy"
+else
+  proxy = "Capsule"
+end
```
ehelms, line 110: Wouldn't Capsule vs Smart Proxy be version dependent? In Katello 3.3+, there are no more "Capsules"

ehelms: OH, looking further below, this would always be 'Smart Proxy' given it talks about the UI.

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+  new_hostname = ARGV[0]
+else
+  puts opt_parser
+  exit
+end
+
+STDOUT.puts("\nA hostname change requires a manual change in the program. " \
+	     "Please log into UI and rename the default #{proxy} (under Infrastructure -> #{proxy}) " \
+             "to #{new_hostname} using both the name and url fields. " \
+             "After this script completes, you will have to re-register any capsules and #{options[:scenario]} clients. " \
```
ehelms: s/capsules/smart proxies/g

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+unless response
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
+
+# Get the hostname from your system
+old_hostname = get_hostname
+
+if is_rhel_version?("6")
```
ehelms, line 144: I suppose if we are backporting this, this makes sense, however we don't support RHEL 6 on 3.3+.

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+rm -rf /etc/candlepin/certs/amqp{,.bak}
+rm -rf /etc/foreman-proxy/*ssl*
+rm -rf /etc/foreman/old-certs
+rm -rf /etc/foreman/*.pem
+rm -rf /var/lib/puppet/ssl
+rm -rf /etc/pki/katello/nssdb
+rm -f /etc/tomcat/keystore
+rm -f /etc/pki/katello/keystore
+cp -r #{options[:ssl_dir]} /root/ssl-build.#{old_hostname}#{timestamp}
+rm -rf /root/ssl-build
```
ehelms: Seems like we are doing something wrong in the installer if oyu have to delete all the final certificates vs. changing them in the installer and it propagating the changes.

ehelms: Should this use the ssl_dir option?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+rm -rf /etc/pki/katello-certs-tools{,.bak}
+rm -rf /etc/candlepin/certs/amqp{,.bak}
+rm -rf /etc/foreman-proxy/*ssl*
+rm -rf /etc/foreman/old-certs
+rm -rf /etc/foreman/*.pem
+rm -rf /var/lib/puppet/ssl
+rm -rf /etc/pki/katello/nssdb
+rm -f /etc/tomcat/keystore
+rm -f /etc/pki/katello/keystore
+cp -r #{options[:ssl_dir]} /root/ssl-build.#{old_hostname}#{timestamp}
```
ehelms: Should this use the ssl_dir option for the backup location?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+rm -rf /root/ssl-build
+`
+STDOUT.puts "backed up #{options[:ssl_dir]} to /root/ssl-build.#{old_hostname}#{timestamp}"
+STDOUT.puts "updating hostname in /mtc/hosts"
```
ehelms: s/mtc/etc

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hosts`
+
+STDOUT.puts "updating hostname in foreman installer scenarios"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml`
+
+STDOUT.puts "removing last_scenario.yml file"
+`rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml`
```
ehelms, line 199: Why is this necessary?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+STDOUT.puts "re-running the installer"
+
+installer = "#{options[:program]}-installer --scenario #{options[:scenario]} -v --certs-regenerate-ca=true --certs-regenerate=true \
+  --foreman-foreman-url https://#{new_hostname} \
+  --foreman-servername #{new_hostname} \
+  --capsule-qpid-router-broker-addr  #{new_hostname}\
+  --certs-ca-common-name #{new_hostname} \
+  --certs-node-fqdn #{new_hostname} \
+  --capsule-parent-fqdn #{new_hostname}\
+  --foreman-proxy-register-in-foreman true"
```
ehelms: If you sed'd the answer file above with the new hostname, why do we need to call each of these options with the new hostname?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+installer << " --disable-system-checks" if options[:system_check]
+STDOUT.puts installer
+installer_output = `#{installer}`
+STDOUT.puts installer_output
```
ehelms, line 210: Does this mean I won't see the installer output in real time? If not, as a user, what do I see when this is running to know my progress?

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+STDOUT.puts "Restarting services"
+`katello-service restart`
```
ehelms: The installer ought to handle this for us?

katello/katello-change-hostname
ehelms: s/capsules/smart proxies or s/capsules/foreman proxies/g

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+STDOUT.puts "
+Hostname change complete!
+
+You will need to reregister any clients or capsules with the #{options[:scenarion]} server
+
+For capsules, you will have to run the following on the #{options[:scenario]} server (replacing <capsule-hostname> with your capsule's hostname):
+capsule-certs-generate --capsule-fqdn '<capsule-hostname>' --certs-tar '~/<capsule-hostname>-certs.tar'
+
+Then follow the output generated by the capsule-certs-generate command
```
ehelms: Sub "foreman-proxy"

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+
+STDOUT.puts "Restarting services"
+`katello-service restart`
+
+STDOUT.puts "
+Hostname change complete!
+
+You will need to reregister any clients or capsules with the #{options[:scenarion]} server
+
+For capsules, you will have to run the following on the #{options[:scenario]} server (replacing <capsule-hostname> with your capsule's hostname):
```
ehelms: Foreman Proxies

katello/katello-change-hostname
```diff
@@ -0,0 +1,190 @@
...
+capsule-certs-generate --capsule-fqdn '<capsule-hostname>' --certs-tar '~/<capsule-hostname>-certs.tar'
```
ehelms: `foreman-proxy-certs-generate` and `foreman-proxy-fqdn`

ehelms (changes requested):

katello/katello-change-hostname
johnpmitsch: I can test it without doing that - would I just call
`#{options[:program]}-installer --scenario #{options[:scenario]} -v --certs-regenerate-ca=true --certs-regenerate=true`?

katello/katello-change-hostname
johnpmitsch: You wont see this in real time, I print the installer command beforehand so the user knows that is running, similar to ansible's output when you run a playbook (you don't see the output from the command until its completed). I'm open to suggestions though

katello/katello-change-hostname
johnpmitsch: thanks, copy and paste error

katello/katello-change-hostname
johnpmitsch: yes, its there to support backporting

katello/katello-change-hostname
johnpmitsch: I had an issue with this when running this script multiple times, I can try to reproduce and share

pgagne:
This script does not update the hostname in the  URLs for kickstart media. It needs to do this. 

When I tried to sync kickstart repos after using the hostname change i ran into https://bugzilla.redhat.com/show_bug.cgi?id=1411530

This should probably be fixed in this script.

johnpmitsch:
@ehelms @jlsherrill @bbuckingham @pgagne @chris1984 updated

johnpmitsch:
The default capsule and installation media are updated in the script using hammer, I removed the unnecessary installer options, and the unnecessary reindex after the installer run.

katello/katello-change-hostname
```diff
@@ -0,0 +1,174 @@
...
+  puts opt_parser
+  exit
+end
+
+# Get the hostname from your system
+old_hostname = get_hostname
+
+STDOUT.puts "Updating default #{proxy}"
+default_capsule_id = `hammer -u #{options[:username]} -p #{options[:password]} capsule list | grep #{old_hostname} | awk '{ print $1 }'`[0]
+capsule_hammer = `hammer -u #{options[:username]} -p #{options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname}`
```
johnpmitsch: The capsule updates successfully, but I get a `Could not update the proxy` message when this command runs. @akofink have you ever seen this?

akofink: @johnpmitsch Here's a potentially related BZ I recently fixed: https://bugzilla.redhat.com/show_bug.cgi?id=1398695

Could you provide the `-d` output with the hammer command?

johnpmitsch: @akofink  Looks like its trying to contact the new url. The funny thing is I can update these values in the UI just fine

```
[vagrant@centos7-katello-3-2 ~]$ hammer -u admin -p changeme -d capsule update --id 1 --url https://foo.example.com:9090 --new-name foo.example.com
[ INFO 2017-01-05 21:47:19 Init] Initialization of Hammer CLI (0.8.0) has started...
[DEBUG 2017-01-05 21:47:19 Init] Running at ruby 2.2.2-p95
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli_config.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_bootdisk.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_docker.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_tasks.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/import.yml has been loaded
[ INFO 2017-01-05 21:47:19 Init] Configuration from the file /etc/hammer/cli.modules.d/katello.yml has been loaded
[DEBUG 2017-01-05 21:47:19 Connection] Registered: foreman
[DEBUG 2017-01-05 21:47:19 API] Global headers: {
        :content_type => "application/json",
              :accept => "application/json;version=2",
    "Accept-Language" => "en"
}
[DEBUG 2017-01-05 21:47:19 API] Follow redirects: never
[ INFO 2017-01-05 21:47:20 Modules] Extension module hammer_cli_foreman (0.8.0) loaded
[ INFO 2017-01-05 21:47:20 Modules] Extension module hammer_cli_foreman_bootdisk (0.1.3) loaded
[ INFO 2017-01-05 21:47:20 Modules] Extension module hammer_cli_import (0.11.3) loaded
[ INFO 2017-01-05 21:47:20 HammerCLI::MainCommand] subcommand organization (HammerCLIForeman::Organization) was removed.
[ INFO 2017-01-05 21:47:20 HammerCLI::MainCommand] subcommand organization (HammerCLIKatello::Organization) was created.
[ INFO 2017-01-05 21:47:20 Modules] Extension module hammer_cli_katello (0.1.3) loaded
[DEBUG 2017-01-05 21:47:20 Init] Using locale 'en'
[DEBUG 2017-01-05 21:47:20 Init] 'mo' files for locale domain 'hammer-cli' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli-0.8.0/locale'
[DEBUG 2017-01-05 21:47:20 Init] 'mo' files for locale domain 'hammer-cli-foreman' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_foreman-0.8.0/locale'
[DEBUG 2017-01-05 21:47:20 Init] 'mo' files for locale domain 'hammer_cli_foreman_docker' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_foreman_docker-0.0.4/locale'
[DEBUG 2017-01-05 21:47:20 Init] 'mo' files for locale domain 'hammer-cli-katello' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_katello-0.1.3/locale'
[ INFO 2017-01-05 21:47:20 HammerCLI::MainCommand] Called with options: {"option_debug"=>true, "option_username"=>"admin", "option_password"=>"***"}
[ INFO 2017-01-05 21:47:20 HammerCLIForeman::SmartProxy] Called with options: {}
[ INFO 2017-01-05 21:47:20 HammerCLIForeman::SmartProxy::UpdateCommand] Called with options: {"option_id"=>"1", "option_url"=>"https://foo.example.com:9090", "option_new_name"=>"foo.example.com"}
[ INFO 2017-01-05 21:47:20 API] Server: https://localhost/
[ INFO 2017-01-05 21:47:20 API] PUT /api/smart_proxies/1
[DEBUG 2017-01-05 21:47:20 API] Params: {
    "smart_proxy" => {
         "url" => "https://foo.example.com:9090",
        "name" => "foo.example.com"
    },
           "name" => "foo.example.com"
}
[DEBUG 2017-01-05 21:47:20 API] Headers: {}
[ERROR 2017-01-05 21:47:23 API] 422 Unprocessable Entity
[DEBUG 2017-01-05 21:47:23 API] {
    "errors" => {
        "base" => [
            [0] "Unable to communicate with the proxy: ERF12-2530 [ProxyAPI::ProxyException]: Unable to detect features ([Errno::EHOSTUNREACH]: No route to host - connect(2) for \"foo.example.com\" port 9090) for proxy https://foo.example.com:9090/features",
            [1] "Please check the proxy is configured and running on the host."
        ]
    }
}
[ERROR 2017-01-05 21:47:23 Exception] <NilClass> nil
Could not update the proxy
```

```
[vagrant@centos7-katello-3-2 ~]$ rpm -qa | grep hammer
tfm-rubygem-hammer_cli_katello-0.1.3-1.el7.noarch
tfm-rubygem-hammer_cli-0.8.0-1.el7.noarch
tfm-rubygem-hammer_cli_foreman_bootdisk-0.1.3-5.el7.noarch
tfm-rubygem-hammer_cli_foreman_tasks-0.0.10-2.fm1_12.el7.noarch
tfm-rubygem-hammer_cli_import-0.11.3-1.el7.noarch
tfm-rubygem-hammer_cli_foreman-0.8.0-1.el7.noarch
tfm-rubygem-hammer_cli_foreman_docker-0.0.4-2.el7.noarch
```

johnpmitsch: @akofink looking at it last night I think I saw the UI uses PATCH instead of PUT

katello/katello-change-hostname
johnpmitsch: @ehelms without this line, I get this message
```
[ERROR 2017-01-05 21:52:09 verbose] You are trying to replace existing installation with different scenario. This may lead to unpredictable states. Use --force to override. You can use --compare-scenarios to see the differences
```
tested on 3.2

katello/katello-change-hostname
johnpmitsch: @akofink 
```
[vagrant@centos7-katello-3-2 ~]$ [vagrant@centos7-katello-3-2 ~]$ hammer -u admin -p changeme -d capsule update --id 1 --url https://foo.example.com:9090 --new-name foo.example.com
-bash: [vagrant@centos7-katello-3-2: command not found
[vagrant@centos7-katello-3-2 ~]$ [ INFO 2017-01-05 21:47:19 Init] Initialization of Hammer CLI (0.8.0) has started...
-bash: syntax error near unexpected token `('
[vagrant@centos7-katello-3-2 ~]$ hammer -u admin -p changeme -vd capsule update --id 1 --url https://foo.example.com:9090 --new-name foo.example.com[ INFO 2017-01-05 21:50:09 Init] Initialization of Hammer CLI (0.8.0) has started...
[DEBUG 2017-01-05 21:50:09 Init] Running at ruby 2.2.2-p95
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli_config.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_bootdisk.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_docker.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/foreman_tasks.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/import.yml has been loaded
[ INFO 2017-01-05 21:50:09 Init] Configuration from the file /etc/hammer/cli.modules.d/katello.yml has been loaded
[DEBUG 2017-01-05 21:50:09 Connection] Registered: foreman
[DEBUG 2017-01-05 21:50:09 API] Global headers: {
        :content_type => "application/json",
              :accept => "application/json;version=2",
    "Accept-Language" => "en"
}
[DEBUG 2017-01-05 21:50:09 API] Follow redirects: never
[ INFO 2017-01-05 21:50:09 API] Server: https://localhost/
[ INFO 2017-01-05 21:50:09 API] GET /apidoc/v2.en.json
[DEBUG 2017-01-05 21:50:09 API] Params: {}
[DEBUG 2017-01-05 21:50:09 API] Headers: {
    :accept => "application/json",
    :params => {}
}
[DEBUG 2017-01-05 21:50:09 API] Cache expired. (default -> 1c2243b268702a6146575fe718fbb6b8)
[DEBUG 2017-01-05 21:50:09 API] Response: Received OK
[DEBUG 2017-01-05 21:50:09 API] Response headers: {
                         :date => "Thu, 05 Jan 2017 21:50:09 GMT",
                       :server => "Apache/2.4.6 (CentOS)",
              :x_frame_options => "SAMEORIGIN",
             :x_xss_protection => "1; mode=block",
       :x_content_type_options => "nosniff",
          :content_disposition => "inline; filename=\"v2.en.json\"",
    :content_transfer_encoding => "binary",
                :cache_control => "private",
              :apipie_checksum => "1c2243b268702a6146575fe718fbb6b8",
                 :x_request_id => "5468d948-5086-470e-93e9-245392a7013b",
                    :x_runtime => "0.032557",
                 :x_powered_by => "Phusion Passenger 4.0.53",
                       :status => "200 OK",
                         :vary => "Accept-Encoding",
             :content_encoding => "gzip",
               :content_length => "42956",
                 :content_type => "text/html; charset=UTF-8"
}
[DEBUG 2017-01-05 21:50:09 API] New apidoc loaded from the server
[ INFO 2017-01-05 21:50:09 Modules] Extension module hammer_cli_foreman (0.8.0) loaded
[ INFO 2017-01-05 21:50:09 Modules] Extension module hammer_cli_foreman_bootdisk (0.1.3) loaded
[ INFO 2017-01-05 21:50:09 Modules] Extension module hammer_cli_import (0.11.3) loaded
[ INFO 2017-01-05 21:50:10 HammerCLI::MainCommand] subcommand organization (HammerCLIForeman::Organization) was removed.
[ INFO 2017-01-05 21:50:10 HammerCLI::MainCommand] subcommand organization (HammerCLIKatello::Organization) was created.
[ INFO 2017-01-05 21:50:10 Modules] Extension module hammer_cli_katello (0.1.3) loaded
[DEBUG 2017-01-05 21:50:10 Init] Using locale 'en'
[DEBUG 2017-01-05 21:50:10 Init] 'mo' files for locale domain 'hammer-cli' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli-0.8.0/locale'
[DEBUG 2017-01-05 21:50:10 Init] 'mo' files for locale domain 'hammer-cli-foreman' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_foreman-0.8.0/locale'
[DEBUG 2017-01-05 21:50:10 Init] 'mo' files for locale domain 'hammer_cli_foreman_docker' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_foreman_docker-0.0.4/locale'
[DEBUG 2017-01-05 21:50:10 Init] 'mo' files for locale domain 'hammer-cli-katello' loaded from '/opt/theforeman/tfm/root/usr/share/gems/gems/hammer_cli_katello-0.1.3/locale'
[ INFO 2017-01-05 21:50:10 HammerCLI::MainCommand] Called with options: {"option_verbose"=>true, "option_debug"=>true, "option_username"=>"admin", "option_password"=>"***"}
[ INFO 2017-01-05 21:50:10 HammerCLIForeman::SmartProxy] Called with options: {}
[ INFO 2017-01-05 21:50:10 HammerCLIForeman::SmartProxy::UpdateCommand] Called with options: {"option_id"=>"1", "option_url"=>"https://foo.example.com:9090", "option_new_name"=>"foo.example.com"}
[ INFO 2017-01-05 21:50:10 API] Server: https://localhost/
[ INFO 2017-01-05 21:50:10 API] PUT /api/smart_proxies/1
[DEBUG 2017-01-05 21:50:10 API] Params: {
    "smart_proxy" => {
         "url" => "https://foo.example.com:9090",
        "name" => "foo.example.com"
    },
           "name" => "foo.example.com"
}
[DEBUG 2017-01-05 21:50:10 API] Headers: {}
[ERROR 2017-01-05 21:50:13 API] 422 Unprocessable Entity
[DEBUG 2017-01-05 21:50:13 API] {
    "errors" => {
        "base" => [
            [0] "Unable to communicate with the proxy: ERF12-2530 [ProxyAPI::ProxyException]: Unable to detect features ([Errno::EHOSTUNREACH]: No route to host - connect(2) for \"foo.example.com\" port 9090) for proxy https://foo.example.com:9090/features",
            [1] "Please check the proxy is configured and running on the host."
        ]
    }
}
[ERROR 2017-01-05 21:50:13 Exception] <NilClass> nil
Could not update the proxy
```

mccun934:
Is this expected, with the errors around -u:

```
# ./katello-change-hostname -u admin -p changeme something.example.com 
Updating default Smart Proxy
Updating installation media paths
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
Could not update the installation media:
  Error: Unrecognised option '-u'
  
  See: 'hammer medium update --help'
updating hostname in /etc/hostname
```

johnpmitsch:
@mccun934 That -u issue is actually with hammer, was this a dev environment?

akofink:
@johnpmitsch You have to put the -u or -p before the subcommand. i.e. `hammer -ufoo -pbar medium update` not `hammer medium update -ufoo -pbar`. This is specified in the hammer usage: `hammer [OPTIONS] SUBCOMMAND [ARG] ...`.

katello/katello-change-hostname
```diff
@@ -0,0 +1,174 @@
...
+
+STDOUT.puts "Updating installation media paths"
+installation_media_ids = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $1 }'`.split("\n")
+installation_media_paths = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $5 }'`.split("\n")
+installation_media_ids.zip(installation_media_paths).each do |id, path|
+  `hammer medium update -u #{options[:username]} -p #{options[:password]} --id #{id} --path #{path.sub! old_hostname, new_hostname}`
```
johnpmitsch: @akofink good catch, I missed that here :smile:

akofink: @johnpmitsch Here's the offending line where the options are after the subcommand.

johnpmitsch:
@mccun934 @bbuckingham @akofink updated to fix the hammer -u issue

katello/katello-change-hostname
```diff
@@ -0,0 +1,174 @@
...
+rm -rf /etc/pki/katello/nssdb
+rm -f /etc/tomcat/keystore
+rm -f /etc/pki/katello/keystore
+cp -r #{options[:ssl_dir]} #{options[:ssl_dir]}.#{old_hostname}#{timestamp}
+rm -rf #{options[:ssl_dir]}
+`
+STDOUT.puts "backed up #{options[:ssl_dir]} to #{options[:ssl_dir]}.#{old_hostname}#{timestamp}"
+STDOUT.puts "updating hostname in /etc/hosts"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/hosts`
+
```
bbuckingham, line 194: On my config, the /etc/hosts is updated, but only the fqdn is updated.  Is that intentional?

e.g. 

before:
```
192.168.121.200 snap-rhel7.example.com snap-rhel7
```
after:
```
192.168.121.200 snap-rhel7-new.example.com snap-rhel7
```

katello/katello-change-hostname
```diff
@@ -0,0 +1,174 @@
...
+installer << " --disable-system-checks" if options[:system_check]
+STDOUT.puts installer
+installer_output = `#{installer}`
+STDOUT.puts installer_output
+
+STDOUT.puts "
+Hostname change complete!
+
+You will need to reregister any #{plural_proxy} or clients with the #{options[:scenario]} server
```
bbuckingham: You may want to uppercase the options[:scenario] here and on 173.  This will make it consistent with the plural_proxy (e.g. Smart Proxy / Capsule)

bbuckingham: Would it be worthwhile to note that clients:
1. need to ensure they have been updated to know about the new server name (e.g. /etc/hosts)
2. need to have katello-ca-consumer-latest.noarch.rpm re-installed before registering
?

chris1984:
@johnpmitsch worked on 3.1 and 3.2 correctly updating /etc/hosts and install media

https://nimbus.everhelper.me/client/notes/share/729445/1kladecxwy9uoyku4m0d

changed from centos7.example.com katello.example.com

Only thing I see that needs to be addressed is the error when it says it cant update the smart proxy. As we talked it might be best to send output to /dev/null until that is fixed or see if there is a work-around.

@bbuckingham what steps did you do to install and change? curious why your /etc/hosts didnt get updated.

katello/katello-change-hostname
johnpmitsch: good points, I can update the `options[:scenario]` to capitalize, and make the message more explicit

katello/katello-change-hostname
johnpmitsch: that isn't intentional, I would have though sed would have taken care of that, I'll see if I can recreate and fix

bbuckingham:
@chris1984, the configuration that I am working with isn't anything that should be drastically different; however, it is a downstream vagrant install from an earlier zstream.  For my test, I chose to take an environment that I have been using for a while and run it through the tool.

katello/katello-change-hostname
johnpmitsch: @bbuckingham I actually take that back! :) Only instances of the FQDN have been changed. I will make a note of this on the completion message

johnpmitsch:
@chris1984 @bbuckingham updated the ending message, and piped the faulty error from hammer to `/dev/null/` (open to suggestions on how to avoid that). I also added a `y/n` confirmation after the command is run since it is making major changes to a system

johnpmitsch:
@ehelms have your comments all been addressed?

bbuckingham:
If a user mistypes an argument they see the output below.  If possible, we may want to only show that when all ran successfully.

```
Hostname change complete!

You will need to re-register any Capsules or clients with the Satellite server. 

When re-registering clients, you will have to reinstall the bootstrap RPM.
Any mention of the sat-snap-rhel7.example.com on your clients will have to change to sat-snap-rhel7-new.example.com (i.e. /etc/hosts)

For Capsules, you will need to regenerate the certs on the Satellite server and reinstall the Capsule

Short hostnames have not been updated, please update those manually.
```

bbuckingham:
Running with --help will include the prompt below.  Would it be possible to omit the confirmation for the --help scenario, since nothing will change?

```
This will modify your system. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]

```

katello/katello-change-hostname
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/ruby
+
+require "socket"
+require "optparse"
+require "highline/import"
+
+DEFAULT_PROGRAM = "foreman"
+DEFAULT_SCENARIO = "katello"
+DEFAULT_SSL_DIR = "/root/ssl-build"
```
sean797: We should get it from the answer file

```
certs:
<snip>
    ssl_build_dir: /root/ssl-build

File: katello/katello-change-hostname [hunk 39]
```diff
@@ -0,0 +1,186 @@
...
+  STDOUT.puts "the new hostname was not changed successfully, exiting script"
+  exit
+end
+
+STDOUT.puts "stopping services"
+`katello-service stop`
+
+STDOUT.puts "deleting old certs"
+`
+rm -rf /etc/pki/katello{,.bak}
```
sean797: Should get this from answer file
```
certs:
<snip>
    pki_dir: /etc/pki/katello

File: katello/katello-change-hostname [hunk 40]
```diff
@@ -0,0 +1,186 @@
...
+rm -rf /etc/pki/katello-certs-tools{,.bak}
+rm -rf /etc/candlepin/certs/amqp{,.bak}
+rm -rf /etc/foreman-proxy/*ssl*
+rm -rf /etc/foreman/old-certs
+rm -rf /etc/foreman/*.pem
+rm -rf /var/lib/puppet/ssl
```
sean797: Should get this from answer file
```
foreman-proxy:
<snip>
    ssldir: /var/lib/puppet/ssl

File: katello/katello-change-hostname [hunk 41]
```diff
@@ -0,0 +1,186 @@
...
+
+STDOUT.puts "stopping services"
+`katello-service stop`
+
+STDOUT.puts "deleting old certs"
+`
+rm -rf /etc/pki/katello{,.bak}
+rm -rf /etc/pki/katello-certs-tools{,.bak}
+rm -rf /etc/candlepin/certs/amqp{,.bak}
+rm -rf /etc/foreman-proxy/*ssl*
```
sean797: Should get this from answer file

```
foreman-proxy:
<snip>
    ssl_ca: /etc/foreman-proxy/ssl_ca.pem
    ssl_cert: /etc/foreman-proxy/ssl_cert.pem
    ssl_key: /etc/foreman-proxy/ssl_key.pem
    foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
    foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
    foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem

File: katello/katello-change-hostname [hunk 42]
```diff
@@ -0,0 +1,186 @@
...
+rm -rf /etc/foreman/old-certs
+rm -rf /etc/foreman/*.pem
```
sean797: Should get this from answer file

```
foreman:
<snip>
    client_ssl_ca: /etc/foreman/proxy_ca.pem
    client_ssl_cert: /etc/foreman/client_cert.pem
    client_ssl_key: /etc/foreman/client_key.pem

**Review by sean797 — CHANGES_REQUESTED** [review]
Added some inline comments, I think we should read some of the paths from the answer file as we allow them to be changed with foreman-installer.

**Comment by bbuckingham** [comment]
Not all of the supported *-installer have the --disable-system-checks option; therefore, if running one one that doesn't the script will fail.  This is an example where user sees the 'Hostname change complete!' message.

**Comment by bbuckingham** [comment]
@johnpmitsch, I completed my review and test.  ACK pending final updates.  That said, I would request a second ACK before merging.

**Comment by chris1984** [comment]
@bbuckingham @johnpmitsch if you want me to do another test run manually let me know with install from ISO etc

**Comment by bbuckingham** [comment]
@johnpmitsch, I attempted to run the script on an el6 server and it appears that the system ruby may not be compatible.

```
# ruby katello-change-hostname.rb --help
katello-change-hostname.rb:5:in `require': no such file to load -- highline/import (LoadError)
	from katello-change-hostname.rb:5

```

As an alternative, the script can be run using the scl that is installed on the box.  E.g.

```
# scl enable tfm "ruby katello-change-hostname.rb --help"
This will modify your system. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]
y
Usage: katello-change-hostname HOSTNAME [OPTIONS]
...
```

**Comment by johnpmitsch** [comment]
@bbuckingham @sean797 @chris1984 updated

File: katello/katello-change-hostname [hunk 43]
```diff
@@ -0,0 +1,225 @@
...
+  new_hostname = ARGV[0]
+else
+  puts opt_parser
+  exit
+end
+
+# Get the hostname from your system
+old_hostname = get_hostname
+
+katello_answers = YAML.load_file('/etc/foreman-installer/scenarios.d/katello-answers.yaml')
```
sean797: Should use `options[:scenario]` not just `katello-answers.yaml`

Also I don't really like `katello_answers` can we call this `scenario_answers`?

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+
+STDOUT.puts "updating hostname in foreman installer scenarios"
+`sed -i -e 's/#{old_hostname}/#{new_hostname}/g' /etc/foreman-installer/scenarios.d/*.yaml`
+
+STDOUT.puts "removing last_scenario.yml file"
+`rm -rf /etc/foreman-installer/scenarios.d/last_scenario.yaml`
+
+STDOUT.puts "re-running the installer"
+
+installer = "#{options[:program]}-installer --scenario #{options[:scenario]} -v --certs-regenerate-ca=true --certs-regenerate=true \
```
sean797, line 203: If my setup has custom certs will this replace them? If so can we do one of the following:
- Notify the user to re install custom certs with `foreman-installer <blah>` when we're done (maybe check and notify them before we do anything as well)
- Allow user to pass new custom certs in
- Someother better solution you can think of ;)

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+
+  opt.on("-p","--password PASSWORD","admin password REQUIRED") do |password|
+    options[:password] = password
+  end
+
+  opt.on("-g","--program PROGRAM","name of the program you are modifying (defaults to #{DEFAULT_PROGRAM})") do |program|
+    options[:program] = program
+  end
+
+  opt.on("-s","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
```
sean797: Small thing but can we have `-S` instead of `-s` ? kafo/foreman-installer accepts capital and its nice to be consistent.

sean797 (changes requested):
@johnpmitsch nearly there I think! A couple of small things, thanks 👍

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+opt_parser.parse!
+
+STDOUT.puts("This will modify your system. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]")
+response = yesno
+unless response
+  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
+  exit
+end
+
+raise 'Must run as root' unless Process.uid == 0
```
ehelms: This might be best at the top, exit out immediately as well as being obvious to find at the top. Sort of like a guard clause in a method.

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+  puts opt_parser
+  exit
+end
+
+if options[:scenario] == "katello"
+  proxy = "Smart Proxy"
+  plural_proxy = "Smart Proxies"
+else
+  proxy = "Capsule"
+  plural_proxy = "Capsules"
```
ehelms: Capsules are no longer a term we use.

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+default_capsule_id = `hammer -u #{options[:username]} -p #{options[:password]} capsule list | grep #{old_hostname} | awk '{ print $1 }'`[0]
+update_default_capsule = "hammer -u #{options[:username]} -p #{options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname}"
+`#{update_default_capsule} 2> /dev/null`
+
+STDOUT.puts "Updating installation media paths"
+installation_media_ids = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $1 }'`.split("\n")
+installation_media_paths = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $5 }'`.split("\n")
+installation_media_ids.zip(installation_media_paths).each do |id, path|
+  `hammer -u #{options[:username]} -p #{options[:password]} medium update --id #{id} --path #{path.sub! old_hostname, new_hostname}`   
+end
```
ehelms, line 142: Shame these are hard coded and not pulled from some runtime configuration.

katello/katello-change-hostname
ehelms: That seems strange since you aren't changing the scenario.

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
...
+STDOUT.puts installer
+installer_output = `#{installer}`
+installer_success = $?.success?
+STDOUT.puts installer_output
+
+if installer_success
+  STDOUT.puts "
+  Hostname change complete!
+  
+  You will need to re-register any #{plural_proxy} or clients with the #{options[:scenario].capitalize} server.
```
ehelms: Could we just say `the server` ? Otherwise this will say 'Katello server' or 'Foreman Proxy Content Server' which its really the Foreman server and Foreman Proxy.

katello/katello-change-hostname
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/ruby
+
+require "socket"
+require "optparse"
+require 'yaml'
+
+DEFAULT_PROGRAM = "foreman"
+DEFAULT_SCENARIO = "katello"
```
sean797: Something like this might be better 

```
File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]

Comment (johnpmitsch) [reply katello/katello-change-hostname]: thanks, was looking for something like that :+1:

**Comment by johnpmitsch** [comment]
@ehelms @sean797 @bbuckingham updated, doing some further testing on 3.2 as well

**Comment by chris1984** [comment]
@johnpmitsch @ehelms did you want me to run some more manual tests?

**Comment by johnpmitsch** [comment]
@chris1984 the more testing the better! If you have time, go for it

**Comment by bbuckingham** [comment]
Output using a downstream build on el6: 
```
# ruby katello-change-hostname.rb sat-snap-rhel6-new.example.com --username admin --password mypassword --program satellite --scenario satellite
katello-change-hostname.rb:32:in `has_disable_system_check_option?': uninitialized constant Gem (NameError)
	from katello-change-hostname.rb:82
```

Note: this isn't an issue on el7.

**Comment by johnpmitsch** [comment]
@bbuckingham on the el6 error, is that using ruby 1.8.3?

**Comment by bbuckingham** [comment]
@johnpmitsch, it is with :  ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]

**Comment by bbuckingham** [comment]
@johnpmitsch, retested on downstream el6 & el7 and no issues observed.  The one pending issue that I am aware of (and perhaps we can cover it with a separate issue/pr) is related to ensuring other configured features (e.g. from answers yaml) that are referencing the old hostname are also updated (e.g. dns, dhcp...).

**Comment by bbuckingham** [comment]
ACK from me.

File: katello/katello-change-hostname [hunk 51]
```diff
@@ -0,0 +1,224 @@
...
+  end
+
+  opt.on("-h","--help","help") do
+    puts opt_parser
+    exit
+  end
+end
+opt_parser.parse!
+
+STDOUT.print("This will modify your system. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and  want to proceed with changing your hostname? [y/n]:")
```
sean797: I think there is a double space between "and" and "want"
``` 
etc...) and  want to proceed

**Review by sean797 — CHANGES_REQUESTED** [review]
Just one small thing :-) Also are you planning on doing the packaging changes in this PR or another?

**Comment by johnpmitsch** [comment]
@sean797 
regarding packaging (answered on IRC)
11:14:20             jomitsch | sean: I am not planning on packaging the changes in that PR            
11:15:15             jomitsch | sean: Its my first time creating a command like that in kpackaging so I
                              | would need some guidance on how to do that

**Comment by johnpmitsch** [comment]
@sean797 updated with packaging changes

File: katello/katello-change-hostname [hunk 52]
```diff
@@ -0,0 +1,224 @@
...
+
+require "socket"
+require "optparse"
+require "rubygems"
+require 'yaml'
+
+raise 'Must run as root' unless Process.uid == 0
+
+DEFAULT_SCENARIO = File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
+DEFAULT_SCENARIO == "katello" ? DEFAULT_PROGRAM = "foreman" : DEFAULT_PROGRAM = DEFAULT_SCENARIO
```
ehelms: Nitpick but this can be confusing to read vs what I would think is the more standard form of the ternary operator assignment:

```
DEFAULT_PROGRAM = DEFAULT_SCENARIO == "katello" ? "foreman" : DEFAULT_SCENARIO
```

katello/katello-change-hostname
```diff
@@ -0,0 +1,224 @@
...
+    exit
+  end
+end
+opt_parser.parse!
+
+# kept as variables for easy changing in backports
+proxy = "Foreman Proxy"
+plural_proxy = "Foreman Proxies"
+
+STDOUT.print("This will modify your system. You will need to re-register any #{ plural_proxy } and #{options[:scenario].capitalize} clients after script completion. #{ plural_proxy } will have to be reinstalled. If you are using custom certificates, you will have to run the #{options[:program]}-installer again with custom certificate options after this script completes. Have you taken the necessary precautions (backups, snapshots, etc...) and want to proceed with changing your hostname? [y/n]:")
```
ehelms: Nitpick - could we cut this line into 120 character chunks so it is easier to read here and in editors?

katello/katello-change-hostname
johnpmitsch: I updated to notify users that they will have to re-install with custom certs and also gave them a warning during the pre-script yes/no check

johnpmitsch:
@ehelms updated, going to do a test run  to double check for any issues

chris1984 (commented):
@johnpmitsch are you still making changes or is this good for another test on 3.1/3.2 el7

johnpmitsch:
@chris1984 Done making changes for the existing comments, so go ahead and test!

ehelms (approved):

sean797 (approved):
Looks good! Thanks @johnpmitsch 👍

chris1984:
Works good, glad it got merged :)

sthirugn:
fyi - logged 5 new issues
http://projects.theforeman.org/issues/18263
http://projects.theforeman.org/issues/18266
http://projects.theforeman.org/issues/18267
http://projects.theforeman.org/issues/18268
http://projects.theforeman.org/issues/18270

PR #324, opened 2016-12-01
refs #17567 - enable use of katello-remove for Foreman proxies

cfouant:


katello/katello-remove
```diff
@@ -1,181 +1,183 @@
...
+]
+
+def remove
+  `katello-service stop`
+
+  puts "Removing RPMs"
+  RPMS.each { | rpm | `yum erase -y "#{rpm}"` }
+
+  puts "Cleaning up configuration files"
+  CONFIG_FILES.each { | file | `rm -rf "#{file}"` if File.exist? file }
```
stbenjam: Maybe `File.rm_rf`instead?

stbenjam:
Looks good to me! I tried it out and seems to work really well.

We need to handle removal of capsule-remove, which is stored in the installer repo: https://github.com/Katello/katello-installer/blob/master/bin/capsule-remove, can you include that in https://github.com/Katello/katello-installer/pull/437 ?

And then in this PR remove it from the katello installer package  (https://github.com/Katello/katello-packaging/blob/master/katello-installer/katello-installer-base.spec#L57 and elsewhere)

katello-installer/katello-installer-base.spec
```diff
@@ -22,39 +22,43 @@ Requires: katello-selinux
...
-mv /etc/capsule-installer/capsule-installer.yaml.rpmsave /etc/capsule-installer/capsule-installer.yaml >/dev/null 2>&1 || :
-mv /etc/katello-installer/answers.katello-installer.yaml.rpmsave /etc/katello-installer/answers.katello-installer.yaml >/dev/null 2>&1 || :
-mv /etc/capsule-installer/answers.capsule-installer.yaml.rpmsave /etc/capsule-installer/answers.capsule-installer.yaml >/dev/null 2>&1 || :
+# See if the currently selected scenario is capsule - and migrate it to be foreman-proxy-content
+if [[ "`/bin/readlink -f /etc/foreman-installer/scenarios.d/last_scenario.yaml`" == "/etc/foreman-installer/scenarios.d/capsule.yaml" ]];
+then
+  mv /etc/foreman-installer/scenarios.d/capsule-answers.yaml /etc/foreman-installer/scenarios.d/foreman-proxy-content-answers.yaml
+  rm /etc/foreman-installer/scenarios.d/last_scenario.yaml
+  ln -s /etc/foreman-installer/scenarios.d/foreman-proxy-content.yaml /etc/foreman-installer/scenarios.d/last_scenario.yaml
+fi
```
stbenjam, line 45: @ehelms  Here's what I was thinking of

katello/katello.spec
```diff
@@ -142,16 +142,16 @@ Useful utilities for debug info collecting
 %files debug
 %{_datadir}/foreman/script/foreman-debug.d/katello-debug.sh
 
-%package capsule
```
stbenjam, line 145: I think we'll need an Obsoletes: katello-capsule on this package

katello/katello.spec
```diff
@@ -142,17 +142,7 @@ Useful utilities for debug info collecting
...
-Summary: Provides a federation of katello services
-BuildArch: noarch
-Requires: findutils
-Requires: foreman-installer-%{name}
```
stbenjam, line 149: I think we still need a package called katello-foreman-proxy - so don't delete all this, just rename capsule to foreman-proxy

cfouant: @stbenjam - ah, that makes a lot more sense :+1:

ehelms:
Would `foreman-proxy-content` instead of `katello-foreman-proxy` be a better fit for the package name? I'm thinking that would provide more consistency from scenario, puppet module, tools and packaging.

stbenjam:
Sure, that's fine, I have no preference about the name as long as capsule's not in it

ehelms:
Once the naming change is made, I'm good with this. Looks like this will fail until the installer changes are fully in place anyway.

katello/katello.spec
```diff
@@ -142,16 +142,17 @@ Useful utilities for debug info collecting
 %files debug
 %{_datadir}/foreman/script/foreman-debug.d/katello-debug.sh
 
-%package capsule
+%package foreman-proxy-content
```
stbenjam: :+1: Looks good to me - this final package though will be named `katello-foreman-proxy-content` - is that what you want @ehelms?

If we want it w/out the katello prefix, use `%package -n foreman-proxy-content` (see https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch10s04.html)

ehelms: I think without the katello prefix is the better option (so use the `-n` option)

stbenjam:
[test]

katello-installer/katello-installer-base.spec
```diff
@@ -113,24 +117,22 @@ install -d -m0755 %{buildroot}/%{_sbindir}
...
 cp -dpR config/katello.migrations %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
 cp -dpR config/katello-devel.yaml %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
 cp -dpR config/katello-devel.migrations %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
-cp -dpR config/capsule.yaml %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
-cp -dpR config/capsule.migrations %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
+cp -dpR config/foreman-proxy-content.yaml %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
+cp -dpR config/foreman-proxy-content.migrations %{buildroot}/%{_sysconfdir}/foreman-installer/scenarios.d
 
-ln -sf %{_datadir}/foreman-installer-katello/bin/capsule-certs-generate %{buildroot}/%{_sbindir}/capsule-certs-generate
+ln -sf %{_datadir}/foreman-installer-katello/bin/foreman-proxy-certs-generate %{buildroot}/%{_sbindir}/capsule-certs-generate
```
stbenjam: Missed a reference here - see the end of the line, should say foreman-proxy-certs-generate

katello/katello.spec
```diff
@@ -142,16 +142,17 @@ Useful utilities for debug info collecting
...
-%package capsule
+%package -n foreman-proxy-content
 Summary: Provides a federation of katello services
 BuildArch: noarch
 Requires: findutils
 Requires: foreman-installer-%{name}
+Obsoletes: katello-capsule
 
-%description capsule
+%description foreman-proxy-content
```
stbenjam: This needs a `-n` too I think

stbenjam (approved):
Green :)

PR #325, opened 2016-12-05
Add katello-client-bootstrap 1.2.1

zjhuntin:
…[1.2.1-1].


ehelms:
This commit shouldn't include a tag from tito. This should just be a bump in the spec manually and the addition of the source to git annex.

PR #327, opened 2016-12-08
Fixes - #16476 - add capsule backup and restore

chris1984:


zjhuntin:
I was able to backup and restore a capsule successfully without any issue. ACK!

chris1984:
Tested on downstream and works correctly to add to Zach's note

stbenjam (changes requested):
We are no longer using the term 'capsule' beginning with Katello 3.3 - please rename this "foreman-proxy-backup"

stbenjam:
Sorry :) Since we're about to merge the changes in the next couple of days it'd be good for this script to start out named right

chris1984:
@stbenjam thanks :) corrected with the new names

stbenjam:
Thanks! One thing, the old versions are still in the PR, and commit message needs updating  too

stbenjam:
One other thought, we've tried to make the tools work regardless of where they are (capsule-remove is going away, katello-remove now does it all.  katello-service handles everything now too) - is there any chance katello-backup could just have a foreman proxy option?

chris1984:
@stbenjam ack I will work on adjusting katello-backup to add a foreman proxy option. It only makes sense based on your last comment if the other tools are going away around capsule. Will recommit after I change and test with katello-backup

ehelms:
@chris1984 planning to revisit this?

chris1984:
@ehelms going to close this one out and work with @cfouant on putting this all into katello-backup/restore like Stephen suggested. Will open another pr either with her or with the suggested changes to katello-backup/restore.

PR #332, opened 2016-12-21
Make devel installer depend on regular installer

dLobatog:
This allows forklift installations to have katello-service, 
katello-certs-check, etc... right off the bat on forklift.

I wasn't sure which branch to PR this to, the original issue https://github.com/Katello/forklift/issues/341 has a link to a fix directly in forklift in case making this dependency would not be appropriate for some reason.

### DIFF
--- a/katello-installer/katello-installer-base.spec
+++ b/katello-installer/katello-installer-base.spec
@@ -60,6 +60,7 @@ foreman-installer --scenario capsule --migrations-only > /dev/null
 Summary:   Installer scenario for Katello development setup from git
 Group:	   Applications/System
 Requires:  %{name} = %{version}-%{release}
+Requires:  foreman-installer-katello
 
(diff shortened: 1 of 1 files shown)


katello-installer/katello-installer-base.spec
```diff
@@ -60,7 +60,7 @@ foreman-installer --scenario capsule --migrations-only > /dev/null
 Summary:   Installer scenario for Katello development setup from git
 Group:	   Applications/System
 Requires:  %{name} = %{version}-%{release}
-
+Requires:  foreman-installer-katello
```
ehelms, line 63: Sorry for the delay getting to this, mind putting a blank line beneath this to separate the description block?

dLobatog:
@ehelms Updated

ehelms (approved):

PR #343, opened 2017-01-05
Update spec files for easier release

jlsherrill:


katello-installer/katello-installer-base.spec
```diff
@@ -2,7 +2,7 @@
 %undefine scl_prefix
 %global scl_ruby /usr/bin/ruby
 
-#global prever .rc1
+#%global prever .rc1
```
ehelms: This will fail AFAIK. The global parameter definition does not respect the comment preceding it which is why we had it this way.

jlsherrill: oh :( did not realize that  we'll see

ehelms:
@jlsherrill fail fail

katello-installer/katello-installer-base.spec
```diff
@@ -2,7 +2,7 @@
 %undefine scl_prefix
 %global scl_ruby /usr/bin/ruby
 
-#global prever .rc1
+# %%global prever .rc1
```
jlsherrill, line 5: @ehelms note, i updated this and supposedly this will work, curious your thoughts.  I think i like this better than just '#global' myself

jlsherrill:
note this will still fail until nightly is fixed

katello-installer/katello-installer-base.spec
ehelms: Is this like a double negative? I thought double `%%` was used to reference a macro or create a macro?

I assume that `#% %global` won't work?

jlsherrill: I got that from here:  http://rpm.org/max-rpm-snapshot/ch-rpm-inside.html

doesn't appear that `#% %global` works according to rpmlint

jlsherrill:
[test]

katello-installer/katello-installer-base.spec
jlsherrill: @ehelms thoughts on this?  I like because I think its more obvious, and it only involves deleting characters rather than having to add characters.

ehelms (approved):

PR #354, opened 2017-01-20
Fixes #18173 - Dont run katello-tracer-upload during install

sean797:
This results in a error because yum has a lock on the RPM DB which katello-tracer-upload needs to run.


jlsherrill:
ACK thanks @sean797 !

PR #356, opened 2017-01-23
Fixes #18198 - collect httpd pulp logs in katello-debug

pmoravec:
Collect /var/log/httpd/pulp-http{s,}_{access,error}_ssl.log*

Fixes #18198

Signed-off-by: Pavel Moravec <pmoravec@redhat.com>

### DIFF
--- a/katello/katello-debug.sh
+++ b/katello/katello-debug.sh
@@ -64,6 +64,8 @@ add_files /var/log/rhsm/*
 add_files /etc/pulp/*.conf
 add_files /etc/httpd/conf.d/pulp.conf
 add_files /etc/pulp/server/plugins.conf.d/nodes/distributor/*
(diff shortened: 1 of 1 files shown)


jlsherrill:
This will only grab the pulp logs on the capsule, on the katello server they exist in foreman-ssl_access_ssl.log
did you mean to grab both?

pmoravec:
foreman-ssl_access_ssl.log is collected by foreman-debug already, see https://github.com/theforeman/foreman/blob/develop/script/foreman-debug#L328 . I often use that file for problem diagnostics already :)

Indeed, Satellite logs there and Capsule to the pulp named. And since it is pulp named, I think katello-debug should collect this one, rather than foreman-debug.

jlsherrill:
right right!  For some reason i didn't remember seeing it. oh well ACK! thanks @pmoravec !

PR #366, opened 2017-01-26
fixes #18277 - adds a metadata file to backup

cfouant:


chris1984 (approved):
Looks good to me

katello/katello-backup
```diff
@@ -123,6 +123,16 @@ else
     '/var/lib/puppet/ssl'
   ]
 
+  puts "Generating metadata ... "
+  metadata = File.new 'metadata', 'w'
+  metadata.puts `cat /etc/redhat-release`
+  metadata.puts "Plugin list: "
+  plugin_list = `foreman-rake plugin:list | grep 'Foreman plugin: '`.lines
```
chris1984, line 130: I will have to save that command for my records, awesome this exists.

stbenjam:
Commit message needs a # before the issue number, otherwise :+1:

katello/katello-backup
cfouant: Right? I looked for so long to find a way... @stbenjam was the man with the answers! Hero of my week!

cfouant:
@stbenjam - fixed the commit message, thanks!

stbenjam (approved):
LGTM

ehelms:
I think it would have been nice if this was a known structured file type (unless I'm missing something). This appears to be free form vs something like YAML or JSON that could be parsed by automation easier.

PR #375, opened 2017-02-16
Make sure the postgres roles exist before doing the data loads

jcpunk:
When migrating to a different host, it is helpful to have the roles created in the restore.

On normal restores these new lines have no impact.

### DIFF
--- a/katello/katello-restore
+++ b/katello/katello-restore
@@ -44,7 +44,9 @@ def restore_databases
     `tar --selinux --overwrite --listed-incremental=/dev/null -xzf pgsql_data.tar.gz -C /`
   else
     `service postgresql start`
+    `runuser - postgres -c "createuser foreman"`
     `runuser - postgres -c "dropdb foreman"`
+    `runuser - postgres -c "createuser candlepin"`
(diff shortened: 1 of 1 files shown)


jlsherrill:
[test]

katello/katello-restore
```diff
@@ -44,7 +44,9 @@ def restore_databases
     `tar --selinux --overwrite --listed-incremental=/dev/null -xzf pgsql_data.tar.gz -C /`
   else
     `service postgresql start`
+    `runuser - postgres -c "createuser foreman"`
```
sean797, line 47: I don't think many people have actually done this but you can change the db_username using foreman-installer. It may be worth parsing `/etc/foreman-installer/scenarios.d/katello-answers.yaml` for the answer and defaulting to `foreman`?

@jlsherrill what do you think? Its also true of the line below for db name. Maybe others as well.

jcpunk: Possibly loading the dump PR #378 would be sufficient to obsolete this whole PR..... I'm not familiar enough with postgresql to be sure.....

jcpunk:
Fixed in other PR

PR #329, opened 2016-12-16
Fixes #18231 - task export is too slow in debug

lzap:
The task export is very slow and it is causing foreman-debug to timeout when
called via sos. This is unacceptable behavior and this mechanism must be pulled
away out of foreman-debug.

As an alternative, I offer fast data dump of relevant tables to have initial
data to start with, we can always ask for the export. This is in SQL DDL for a
reason - this way we can import the data into database (sqlite3 should work
too) for visual investigation if needed. From my experience, opening the SQL
file with less is good enough for basic knowledge (does task exist, what was
its result, when it finished). The only drawback is hierarchy.

@pmoravec @ehelms @inecas

PR #377, opened 2017-02-20
Fixes #18571 - Address the use of known directories in collecting the…

bkearney:
… debug data

### DIFF
--- a/katello/katello-debug.sh
+++ b/katello/katello-debug.sh
@@ -126,9 +126,16 @@ add_cmd "rpm -qa | grep qpid" "qpid-rpm-qa"
 add_cmd "mongo pulp_database --eval \"db.reserved_resources.find().pretty().shellPrint()\"" "mongo-reserved_resources"
 add_cmd "mongo pulp_database --eval \"db.task_status.find().pretty().shellPrint()\"" "mongo-task_status"
 
-echo "db.task_status.find({state:{\$ne: \"finished\"}}).pretty().shellPrint()" > /tmp/pulp_running_tasks.js
+TEMP_DIR=`mktemp -d`
+cleanup() {
+  rm $TEMP_DIR/pulp_running_tasks.js
+  rmdir $TEMP_DIR
+}
+trap "cleanup" SIGHUP SIGINT SIGTERM EXIT
 
-add_cmd "mongo pulp_database /tmp/pulp_running_tasks.js" "pulp-running_tasks"
+echo "db.task_status.find({state:{\$ne: \"finished\"}}).pretty().shellPrint()" > $TEMP_DIR/pulp_running_tasks.js
+
+add_cmd "mongo pulp_database $TEMP_DIR/pulp_running_tasks.js" "pulp-running_tasks"
 
 add_cmd "hammer ping" "hammer-ping"
 


bkearney:
@ehelms why was this removed from the issue?

ehelms:
I don't follow the question. What is the this being removed?

bkearney:
@ehelms on http://projects.theforeman.org/issues/18571 this pull request was removed.

ehelms:
Ohh, dang, it removed it when I tried to do a mass move of the issues. I added it back.

lzap:
Oh I was going to fix this in my other patch, I will remove it from there: https://github.com/Katello/katello-packaging/pull/329

lzap:
@bkearney I think the whole idea of TEMP files is bad, the target directory is stored in `$DIR` variable, just use it: `... > "$DIR/whatever_export.log" and get rid of TEMP file completely.

lzap:
Oh I see it's really temporary input for mongo client, can't you use STDIN and pipe these guys together, so it's just one add_cmd statement? If not and you want to stick with temp files, please delete them nicely:

```bash
 cleanup() {
  rm -f $TMP_PULP_THING
}
trap "cleanup" SIGHUP SIGINT SIGTERM EXIT
```

ehelms:
Please rebase @bkearney

bkearney:
@lzap added your cleanup
@ehelms rebased.

bkearney:
@ehelms I have played this go around before, where should I put the ticket?

lzap:
Well it's much better to put the function and trap signal at the top before the command that peforms on TMP so there is no window when this can be canceled.

bkearney:
@lzap moved the location. Although, as a tmp file it would get cleaned up.

katello/katello-debug.sh
```diff
@@ -126,9 +126,19 @@ add_cmd "rpm -qa | grep qpid" "qpid-rpm-qa"
...
+
+echo "db.task_status.find({state:{\$ne: \"finished\"}}).pretty().shellPrint()" > $TEMP_DIR/pulp_running_tasks.js
+
+add_cmd "mongo pulp_database $TEMP_DIR/pulp_running_tasks.js" "pulp-running_tasks"
+
+cleanup() {
+  rm $TEMP_DIR/pulp_running_tasks.js
+  rmdir $TEMP_DIR
+}
+trap "cleanup" SIGHUP SIGINT SIGTERM EXIT
```
lzap, line 134: This line needs to be above the `echo` otherwise you can still experience uncleaned temp file.

katello/katello-debug.sh
```diff
@@ -126,9 +126,19 @@ add_cmd "rpm -qa | grep qpid" "qpid-rpm-qa"
...
 
-echo "db.task_status.find({state:{\$ne: \"finished\"}}).pretty().shellPrint()" > /tmp/pulp_running_tasks.js
+TEMP_DIR=`mktemp -d`
+
+echo "db.task_status.find({state:{\$ne: \"finished\"}}).pretty().shellPrint()" > $TEMP_DIR/pulp_running_tasks.js
+
+add_cmd "mongo pulp_database $TEMP_DIR/pulp_running_tasks.js" "pulp-running_tasks"
+
+cleanup() {
+  rm $TEMP_DIR/pulp_running_tasks.js
```
lzap: I'd simply remove whole directory recursively `rm -rf`

katello/katello-debug.sh
bkearney: @lzap fixed and pushed

bkearney:
@lzap the WOC is a problem with the category. The latest push has your changes.

lzap:
What a safe deletion code, good work :-) I like that, have you already wiped whole file system? ;-) Thanks.

lzap:
WFM RTM but I don't have push rights, @ehelms ? Thanks @bkearney

PR #379, opened 2017-02-20
Fixes #18333 katello-backup returns 0 exit code even when failing

jcpunk:
Checking the success of the various backup operations.  Aborts the backup with a hopefully useful error message on failure.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -44,14 +44,22 @@ optparse = OptionParser.new do |opts|
   @dir = ARGV[0].dup
 end
 
+def run_cmd(command)
+  `#{command}`
+  if $?.exitstatus != 0
+    STDERR.puts "Failed '#{command}'"
+    exit(-1)
+  end
+end
+
 def backup_db_online
-  puts "Backing up postgres online schema db... "
-  `runuser - postgres -c "pg_dump -Fc foreman > #{@dir}/foreman.dump"`
-  `runuser - postgres -c "pg_dump -Fc candlepin > #{@dir}/candlepin.dump"`
+  puts "Backing up postgres online schema... "
+  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
+  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
   puts "Done."
 
-  puts "Backing up mongo online schema db... "
-  `mongodump --host localhost --out mongo_dump`
+  puts "Backing up mongo online schema... "
+  run_cmd('mongodump --host localhost --out mongo_dump')
   puts "Done."
 end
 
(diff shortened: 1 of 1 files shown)


stbenjam:
I'd extract all the logic to a method that executes a command, and then fails with a message if the command doesn't succeed instead of having the same block of code repeated throughout the program.

jcpunk:
Something like this last commit?

ehelms:
@stbenjam could you take another look at this, thanks!

cfouant:
@stbenjam ping, could you re-review pls? Thanks!

cfouant:
@jcpunk could you rebase (you might need to do it again, depending on which of your PRs merges next). Thanks!

jcpunk:
:) Can do, rebased

stbenjam:
[test]

katello/katello-backup
```diff
@@ -40,14 +40,22 @@ optparse = OptionParser.new do |opts|
   @dir = ARGV[0].dup
 end
 
+def run_cmd(command)
+  `#{command}`
+  if $?.exitstatus != 0
+    puts "Failed '#{command}'"
```
stbenjam: Overall the change looks good to me, I'd prefer if we did `STDERR.puts` here instead (and everywhere else we print an error and exit).

Some sysadmins have cron send stdout to /dev/null, but keep stderr so they get a mail from cron when the job fails that has the error message in it.

jcpunk:
Updated to use `STDERR.puts`

stbenjam:
👍 
Now just needs a rebase since https://github.com/Katello/katello-packaging/pull/389 was just merged

jcpunk:
rebased

ehelms:
I am taking @stbenjam thumbs up as approval since this looks good I am going to merge.

cfouant:
Hey guys - this actually breaks katello-backup. Any way we can revert this merge?

katello/katello-backup
```diff
@@ -44,14 +44,22 @@ optparse = OptionParser.new do |opts|
   @dir = ARGV[0].dup
 end
 
+def run_cmd(command)
+  `#{command}`
+  if $?.exitstatus != 0
```
cfouant, line 49: The config file backup will generally return exitstatus 2, as the file systems are different for puppet 3 vs puppet 4, which means some files that are called in the compression don't exist, based on which puppet is being used.

jcpunk: If you've got error snippets, I can make it smarter.  I'd rather fix the check than defeat it.

Something like -> https://github.com/jcpunk/katello-packaging/blob/puppet_conf_3_vs_4/katello/katello-backup ?

ehelms: Your branch has what @cfouant and I were discussing as a possible alternative form.

ehelms: @jcpunk can you open a PR with your proposed update?

jcpunk: Done -> #405

PR #383, opened 2017-02-22
Modified katello-service to support enable/disable for systemd

fftux:
Changes in PR #306 lay ground work for enable/disable support.  However, when enable/disable is invoked on a systemd system, it parses the command as "service-wait #{service} #{action}".  Thus, I modified the manage_services method to utilize systemctl and chkconfig where necessary.  It's crude, but it works.  Not a dev.

### DIFF
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -91,10 +91,9 @@ def manage_services(action)
         puts "#{COMMAND} #{service} #{action}"
         puts `#{COMMAND} #{service} #{action}`
       end
-    end
     failures << service unless $?.success?
+    end
   end
-end
 
   if failures.empty?
     puts "Success!"
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -94,6 +94,7 @@ def manage_services(action)
     end
     failures << service unless $?.success?
   end
+end
 
   if failures.empty?
     puts "Success!"
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -36,10 +36,6 @@ OptionParser.new do |opts|
     @options[:only] = only
   end
 
-  opts.on("-v", "--verbose", "To be used with status, to increase verbosity") do |v|
-    @options[:verbose] = v
-  end
-
   opts.parse!
 
   if ARGV.length == 0 || ARGV.length != 1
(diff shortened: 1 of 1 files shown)


katello/katello-service
```diff
@@ -78,7 +78,19 @@ def manage_services(action)
...
+        puts "systemctl #{action} #{service}.service"
+        puts `systemctl #{action} #{service}.service`
+      else
+        case action
+        when 'enable', 'disable'
+          puts "chkconfig #{service} on"
+          puts `chkconfig #{service} on`
+        else
+          puts "#{COMMAND} #{service} #{action}"
+          puts `#{COMMAND} #{service} #{action}`
```
ehelms: Is the double puts intended for each action?

ehelms:
Thanks for the PR @chopskxw ! This looks like it would change not only disable/enable but other commands as well? I think we would need to test whether moving away from service-wait for some services would work for us. The original intent of service-wait was due to system services deeming a service "up" or "running" even though said service was not reachable yet (e.g. waiting on a port to become available to talk to or a the webserver to be up). This was written many years ago, and systemd may properly handle this now.

fftux:
I can back out the use of systemctl here, and rely on chkconfig to do the work.  However, this will likely become antiquated soon rather than later.  Obviously there are a few different ways to achieve the end goal, this just seemed like the path of least resistance.  Will be happy to make some mods, if necessary.

ehelms (changes requested):

katello/katello-service
fftux: Without the double puts, there is not output of what is going on.  Just a blank line for each command, followed by the "Success" or failures.

ehelms:
@chopskxw What if we started with just checking for enable/disable and use systemd ? And continue using service-wait for other service actions? 

The next step would be to look at what service-wait (https://github.com/Katello/katello-packaging/blob/master/katello/service-wait) is doing and decide whether it has any antiquated ways of checking if a service is started or not given systemd on EL7 to begin moving away from it.

fftux:
@ehelms reorganized method to check for enable/disable action first, and only run systemctl when for these functions.

katello/katello-service
```diff
@@ -78,8 +78,20 @@ def manage_services(action)
...
+          puts `systemctl #{action} #{service}.service`
+        else
+          puts "chkconfig #{service} on"
+          puts `chkconfig #{service} on`
+        end
+      else
+        puts "#{COMMAND} #{service} #{action}"
+        puts `#{COMMAND} #{service} #{action}`
+      end
+    failures << service unless $?.success?
```
ehelms: Looks like this should be indented to appear inside the if-else

ehelms:
Seems reasonable -- I will test these changes locally.

ehelms:
Looks and works good, Thanks @chopskxw !

PR #388, opened 2017-02-27
Refs #18247 - Show initial org in proxy certs gen

chris1984:


chris1984:
@stbenjam is this failing because of my pr or something else going on with the rpms? Also do you want me to file a new issue for this to keep foreman bot happy or just ignore him

stbenjam:
My bad, if you go down the jenkins rabbit hole it'll eventually take you to the Koji job, it's failing because foreman-proxy-certs-generate is in bin not config. 

http://koji.katello.org/koji/getfile?taskID=606272&name=build.log

My comment was wrong, https://github.com/Katello/katello-installer/pull/472#discussion_r103268646, use ` bin/foreman-proxy-certs-generate ` for the path

chris1984:
oh I see what its doing now, dumb part on me too, we are seding the config path then pointing to the file location, since it lives in bin we need to point there. Sorry this even came up

stbenjam:
You can ignore foreman bot about the issue #.

chris1984:
@stbenjam ack all set with the correct commit, me-- for not seeing the obvious.

stbenjam:
ACK, Thanks @chris1984

PR #389, opened 2017-02-27
Run the tar compression in parallel

jcpunk:
I believe this should allow the compression of the postgres/mongo tar files to happen in parallel and thus a bit faster.


cfouant:
[test]

cfouant:
ACK

PR #390, opened 2017-03-01
Update katello-repos.spec so that repo files are config files

Klaas-:
If you update the .repo files, for example to disable repositories because you have them mirrored locally (self registered katello) an update will simply overwrite the changes. if you have them marked as config files you get a warning about the rpmnew/rpmsave files while updating.

someone please test it, I don't have an rpmbuild environment handy at the moment :)


ehelms:
[test]

ehelms:
@Klaas- Thanks! This change makes sense, I have set it to kick off a test build.

PR #397, opened 2017-03-20
Refs #18403 - Regenerate ueber certs after hostname change

johnpmitsch:
Depends on https://github.com/Katello/katello/pull/6680

Since we regenerate the CA, we need to regenerate the
ueber cert for each Organization.

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -214,6 +214,9 @@ installer_output = `#{installer}`
 installer_success = $?.success?
 STDOUT.puts installer_output
 
+STDOUT.puts "Regenerating ueber certs"
+`foreman-rake katello:regenerate_ueber_certs`
+
 if installer_success
   STDOUT.puts "
   Hostname change complete!


katello/katello-change-hostname
```diff
@@ -214,6 +214,9 @@ installer_output = `#{installer}`
 installer_success = $?.success?
 STDOUT.puts installer_output
 
+STDOUT.puts "Regenerating ueber certs"
+`foreman-rake katello:regenerate_ueber_certs`
```
ehelms, line 218: Is the hostname change script expected to run on a proxy?

johnpmitsch:
when I ran on 3.3 with https://github.com/Katello/katello/pull/6680 added to katello, ueber certs were successfully changed after a hostname change run :+1:

katello/katello-change-hostname
johnpmitsch: @ehelms the hostname change script is only for the main katello server

ehelms:
Thanks @johnpmitsch !

jlsherrill:
I know this was just merged, but would it make more sense to regenerate this when the installer runs with --certs-regenerate-ca=true ? 

Just had a user run that by hand when switching between custom certs and non-custom certs and they hit this problem

johnpmitsch:
@jlsherrill the hostname change script runs with --certs-regenerate-ca=true passed to the installer [here](https://git.io/vyj7n) Is that what you mean?

jlsherrill:
right, but the user may run the installer with --certs-regenerate-ca=true without running the hostname script.  

I'm proposing moving the calling of this rake task into the installer, to call be called if the installer is run with --certs-regenerate-ca=true

johnpmitsch:
@jlsherrill ah, okay, ya I think that is a good idea, maybe the ueber cert not regenerating is a cause of some of the custom cert issues we see

PR #398, opened 2017-03-22
Fixes #18965 - exclude unnecessary services

lzap:
The script does not touch these, goferd is leaking memory when connection is lost, let's not poke it when not necessary.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -7,6 +7,8 @@ require 'date'
 @options = {}
 @dir = nil
 
+EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
+
 optparse = OptionParser.new do |opts|
   opts.banner = "Usage: katello-backup /path/to/dir [options]\n eg: $ katello-backup /tmp/katello-backup"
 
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -7,6 +7,8 @@ require 'date'
 @options = {}
 @dir = nil
 
+EXCLUDED="--exclude goferd,foreman-proxy,squid,qdrouterd,qpidd"
```
sean797: Is `smart_proxy_dynflow_core` required to be taken down as well?

cfouant: Verified that smart_proxy_dynflow_core can also be excluded. Could you add it @lzap ? Thanks!

lzap: Super nice!

lzap:
Amended, yay!

sean797 (approved):
Thanks @lzap 🍾

cfouant:
ACK @ehelms - can we have a merge? Thanks @lzap !

ehelms:
This will need to be rebased due to recent other merges.

lzap:
Done!

ehelms:
Thanks @lzap

PR #399, opened 2017-03-22
Fixes #18242 - Omit katello-export directory in katello-backup

johnpmitsch:
This directory doesn't need to be included in the backup, we
can save space by not syncing it.


cfouant:
ACK works great, thanks @johnpmitsch !

PR #400, opened 2017-03-23
Fixes #18270 - Restart puppet services after hostname change

johnpmitsch:


johnpmitsch:
@akofink 

To test:
- Spin up a 3.3 box
- Run this (as root):
```
yum install -y git
cd ~
 git fetch origin pull/400/head:pr400
 git checkout pr400
```
- setup a product with a puppet repo and upload some puppet modules to that repo
- create a couple lifecycle envs
- create a CV and add the puppet modules
- publish and promote a CVV through the LEs
- run hostname change
`./katello-change-hostname foo.example.com -u admin -p changeme -d`
- publish and promote a new CVV (in the same CV) through lifecycle environments

If you are able to promote a content view version with puppet modules in it after a hostname change, then bug fixed! :smile: let me know if you have any questions

katello/katello-change-hostname
```diff
@@ -217,6 +217,10 @@ STDOUT.puts installer_output
 STDOUT.puts "Regenerating ueber certs"
 `foreman-rake katello:regenerate_ueber_certs`
 
+STDOUT.puts "Restarting puppet services"
+`service puppet restart`
+`service puppetserver restart`
```
sean797: systemctl ? were not supporting el6 now ;) 
Also is puppetserver running on every box now? I though we run it with Passenger, at least for Puppet 3

sean797 (changes requested):

katello/katello-change-hostname
johnpmitsch: @sean797 That is true, but this script is given out to use on older versions of Katello, so I used `service` to still support them.

I'm unsure about that. I had to restart puppetserver on 3.3, maybe @ehelms knows more

akofink: @johnpmitsch you might want to use `/sbin/service ...`. I just fixed a bug with katello-backup where the system alias in EL7 doesn't work properly if you run it from a script (i.e. a cron job that runs katello-backup). It's worth considering. https://bugzilla.redhat.com/show_bug.cgi?id=1330757

ehelms:
Looks good pending @akofink comment

akofink:
@johnpmitsch 

I'm not sure if these errors are acceptable or not. I used the centos7-katello-3.3 box on forklift:

While running `katello-change-hostname`:

```
Regenerating ueber certs
rake aborted!
Don't know how to build task 'katello:regenerate_ueber_certs'
```

Content view publishing worked fine, but promoting that version to the next lifecycle environment, I got:

```
ERF12-4115 [ProxyAPI::ProxyException]: Unable to get classes from Puppet for KT_Default_Organization_dev_cv1_2 ([RestClient::NotAcceptable]: 406 Not Acceptable) for proxy https://foo.example.com:9090/puppet
```

Here's the task export: [task-export-1490618355.tar.gz](https://github.com/Katello/katello-packaging/files/872459/task-export-1490618355.tar.gz)

johnpmitsch:
@akofink I forgot to mention that you have to comment out [this line](https://git.io/vSttM) as that rake task is not in 3.3 (an alternative is to pull in  the rake task as a patch https://github.com/Katello/katello/commit/b0a41ac9bc0c2a5ae780d8d544a33f3cb68e5300)

After you do that, you can re-run katello-change-hostname and try the CV promotion again. Sorry for the confusion

johnpmitsch:
@akofink updated to use `/sbin/service`

ehelms:
Looks good to me, will let @akofink double check

akofink:
After commenting out the line mentioned above, publish and promotion works as expected after the katello-change-hostname. ACK!

ehelms:
Thanks @johnpmitsch !

PR #402, opened 2017-03-24
Fixes #19018 - katello-service sends logger messages

lzap:
As easy as that, this will be much cleaner in system log when user attempted to restart services.

### DIFF
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -75,6 +75,7 @@ end
 
 def manage_services(action)
   failures = []
+  `logger -t #{$0} -p daemon '*** #{action} initiated ***'`
 
   services_by_priority(action == 'stop').each do |service|
     if service_exists?(service)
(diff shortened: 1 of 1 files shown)


sean797:
I don't understand why anyone would want this? Systemd already logs this data, e.g.

```
Mar 24 12:38:34 centos7-foreman-1-14.example.com systemd[1]: Stopped The Apache HTTP Server.
Mar 24 12:38:50 centos7-foreman-1-14.example.com systemd[1]: Starting The Apache HTTP Server...
Mar 24 12:38:50 centos7-foreman-1-14.example.com systemd[1]: Started The Apache HTTP Server.
```

lzap:
Yes, totally agree but I encountered a customer report which was unclear - they were doing stop/start/stop/stop and systemd was reporting starting/stopping services, it was not clear. All this does it better transparency, but if you don't like it, just close.

ehelms:
I am unfamiliar with the `logger` function. Is that available on EL6 and EL7? I think @lzap point is valid for distinguishing the use of `katello-service`. Do you have sample output @lzap of when this is used?

lzap:
Well it's everywhere, `man 1 logger` even in ancient RHELs. Part of util-linux. And it's like:

```
Mar 24 12:38:34 xxx.example.com katello-service[16233]: *** stop initiated ***
```

ehelms:
Looks good to me -- @sean797 any other thoughts?

sean797 (approved):
@lzap maybe they had messed around with their syslog config, but i guess it could be useful 👍

lzap:
Thanks guys.

PR #403, opened 2017-03-27
Fixes #19038 - Start goferd on install

ehelms:
@jortel @zjhuntin @jlsherrill  Could you please check this to ensure it does not interfere with the original client update workflow that changed this area of code?

The previous work introduced a regression whereby installing the katello-agent package would not start the agent automatically.


zjhuntin:
I have tested this and made sure that it does start goferd on install. It also does continue to properly update from the webui! :tada: :+1:

jlsherrill:
ACK from me then

PR #405, opened 2017-03-27
Fixes #18333 run_cmd takes exit status args

jcpunk:
This is a followup patch for #18333.

The if condition is a bit ugly, but I believe it gets the job done.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -46,9 +46,9 @@ optparse = OptionParser.new do |opts|
   @dir = ARGV[0].dup
 end
 
-def run_cmd(command)
+def run_cmd(command, exit_codes=[0])
   `#{command}`
-  if $?.exitstatus != 0
+  unless exit_codes.include?($?.exitstatus)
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -46,9 +46,9 @@ optparse = OptionParser.new do |opts|
   @dir = ARGV[0].dup
 end
 
-def run_cmd(command)
+def run_cmd(command, exit_codes=[0])
   `#{command}`
-  if $?.exitstatus != 0
+  if not exit_codes.include?($?.exitstatus)
```
cfouant: Let's do `unless exit_codes.include?($?.exitstatus)` rather than `if not`. Thanks @jcpunk I don't think it's too bad, I had drawn up something similar.

jcpunk: Can do, I'll have a fixed one up shortly.

cfouant:
[test]

cfouant:
ACK, thanks @jcpunk ! @ehelms ready for merge.

PR #406, opened 2017-03-28
Fixes #19054 - Use a timestamp without special characters in kbackup

johnpmitsch:
These special characters can cause issues for programs, the
timestamp will now read 20170328172811, allowing the files
to still be sorted by date.


cfouant:
ACK - thanks @johnpmitsch !

PR #414, opened 2017-04-06
fixes #18329 - adds LVM snapshots to backup

cfouant:
To test:
Set up a katello instance on a LVM filesystem (cannot use the Vagrant set up). You may create partitions separately, such as mounting /var/lib/pulp separate from root. The volume group must have enough space to hold a snapshot, which by default would be about 6G.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -3,24 +3,52 @@
 require 'optparse'
 require 'fileutils'
 require 'date'
-require "yaml"
+require 'yaml'
+require 'find'
+
+EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
+DATABASES = ['pulp', 'pgsql', 'mongodb']
+CONFIGS = [
+  '/etc/candlepin',
+  '/etc/foreman',
+  '/etc/foreman-proxy',
+  '/etc/hammer',
+  '/etc/httpd',
+  '/etc/foreman-installer',
+  '/etc/pki/katello',
+  '/etc/pki/katello-certs-tools',
+  '/etc/pki/pulp',
+  '/etc/pulp',
+  '/etc/puppet',
+  '/etc/puppetlabs',
+  '/etc/qpid',
+  '/etc/qpid-dispatch',
+  '/etc/sysconfig/tomcat*',
+  '/etc/tomcat*',
+  '/root/ssl-build',
+  '/var/lib/candlepin',
+  '/var/www/html/pub',
+  '/opt/puppetlabs/puppet/cache/foreman_cache_data',
+  '/opt/puppetlabs/puppet/ssl',
+  '/var/lib/puppet/foreman_cache_data',
+  '/var/lib/puppet/ssl'
+]
 
 @options = {}
 @dir = nil
-
-EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
+@databases = DATABASES.dup
 
 optparse = OptionParser.new do |opts|
   opts.banner = "Usage: katello-backup /path/to/dir [options]\n eg: $ katello-backup /tmp/katello-backup"
 
   opts.on("--skip-pulp-content", "Create backup without Pulp content for debugging only") do |config_only|
     @options[:config_only] = config_only
+    @databases.delete 'pulp'
   end
 
-  opts.on("--incremental [PREVIOUS_BACKUP_DIR]", String, "Backup changes since previous backup") do |dir_path|
+  opts.on("--incremental PREVIOUS_BACKUP_DIR", String, "Backup changes since previous backup") do |dir_path|
     opts.abort("Please specify the previous backup directory.") unless dir_path
     if File.directory?(dir_path)
-      dir_path.chop! if dir_path.end_with? "/"
       @options[:incremental] = dir_path
     else
       opts.abort("Previous backup directory does not exist: #{dir_path}")
@@ -33,6 +61,18 @@ optparse = OptionParser.new do |opts|
 
   opts.on("--logical-db-backup", "Also dump full database schema during offline backup") do |logical|
     @options[:logical_backup] = logical
+  end
+
+  opts.on("--snapshot", "Use snapshots of the databases to create backup") do |snapshot|
+    @options[:snapshot] = snapshot
+  end
+
+  opts.on("--snapshot-mount-dir SNAPSHOT_MOUNT_LOCATION", String, "Override default directory (/var/snap/) where the snapshots will be mounted") do |mount_dir|
+    @options[:snapshot_mount] = mount_dir
+  end
+
+  opts.on("--snapshot-size SNAPSHOT_BLOCK_DEVICE_SIZE", String, "Override default block size (2G)") do |size|
+    @options[:snapshot_size] = size
   end
 
   opts.parse!
@@ -48,11 +88,12 @@ optparse = OptionParser.new do |opts|
 end
 
 def run_cmd(command, exit_codes=[0])
-  `#{command}`
+  result = `#{command}`
   unless exit_codes.include?($?.exitstatus)
     STDERR.puts "Failed '#{command}'"
     exit(-1)
   end
+  result
 end
 
 def create_directories(directory)
@@ -63,59 +104,128 @@ def create_directories(directory)
   FileUtils.chmod_R 0770, directory
 end
 
-def backup_db_online
-  puts "Backing up postgres online schema... "
-  run_cmd("runuser - postgres -c 'pg_dumpall -g > #{@dir}/pg_globals.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
-  puts "Done."
-
-  puts "Backing up mongo online schema... "
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def snapshot_backup
+  @mountdir = @options[:snapshot_mount] || "/var/snap"
+  @snapsize = @options[:snapshot_size] || "2G"
+  FileUtils.mkdir_p @mountdir
+  run_cmd("katello-service stop #{EXCLUDED}")
+  create_and_mount_snapshots
+  run_cmd("katello-service start #{EXCLUDED}")
+  backup_and_destroy_snapshots
 end
 
-def backup_pulp_online
-  puts "Backing up Pulp data... "
-  matching = false
- 
-  until matching
-    checksum1 = `find /var/lib/pulp -printf '%T@\n' | md5sum`
-    create_pulp_data_tar
-    checksum2 = `find /var/lib/pulp -printf '%T@\n' | md5sum`
-    matching = (checksum1 == checksum2)
+def create_and_mount_snapshots
+  @databases.each do |database|
+    puts "Creating #{database} snapshot"
+    lv_info = get_lv_info(database)
+    run_cmd("lvcreate -n#{database}-snap -L#{@snapsize} -s #{lv_info[0]}")
+
+    mount_location = File.join(@mountdir, database)
+    FileUtils.mkdir_p mount_location
+    puts "Mounting #{database} snapshot on #{mount_location}"
+    options = lv_info[1] == 'xfs' ? "-onouuid,ro" : "-oro"
+    run_cmd("mount #{get_snapshot_location(database)} #{mount_location} #{options}")
   end
-  puts "Done."
 end
 
-def backup_db_offline
-  puts "Backing up postgres db... "
-  run_cmd('tar --selinux --create --file=pgsql_data.tar --listed-incremental=.postgres.snar /var/lib/pgsql/data/')
-  puts "Done."
-
-  puts "Backing up mongo db... "
-  run_cmd('tar --selinux --create --file=mongo_data.tar --listed-incremental=.mongo.snar --exclude=mongod.lock /var/lib/mongodb/')
-  puts "Done."
+def get_snapshot_location(database)
+  run_cmd("lvs --noheadings -o lv_path -S lv_name=#{database}-snap").strip
 end
 
-def backup_pulp_offline
-  puts "Backing up Pulp data... "
-  create_pulp_data_tar
-  puts "Done."
+def get_lv_info(database)
+  target = database
+  target = File.join(database, 'data') if target == 'pgsql'
+  run_cmd("findmnt -n --target #{File.join('/var/lib', target)} -o SOURCE,FSTYPE").split
 end
 
-def create_pulp_data_tar
-  run_cmd('tar --selinux --create --file=pulp_data.tar --exclude=/var/lib/pulp/katello-export --listed-incremental=.pulp.snar /var/lib/pulp/')
+def get_base_directory(database)
+  target = ""
+  case database
+  when 'pulp'
+    target = '0005_puppet_module_name_change.txt'
+  when 'pgsql'
+    target = 'postgresql.conf'
+  when 'mongodb'
+    target = 'mongod.lock'
+  end
+
+  result = nil
+  Find.find(File.join(@mountdir, database)) do |path|
+    result = File.dirname(path) if File.basename(path) == target
+  end
+  result
 end
 
-def compress_files
-  psql = spawn('gzip', 'pgsql_data.tar')
-  mongo = spawn('gzip', 'mongo_data.tar')
-  Process.wait(psql)
-  Process.wait(mongo)
+def backup_and_destroy_snapshots
+  @databases.each do |database|
+    basedir = get_base_directory(database)
+    puts "Backing up #{database} from #{basedir}"
+    offline_backup(database, basedir)
+
+    puts "Unmounting #{database} snapshot"
+    run_cmd("umount #{File.join(@mountdir, database)}")
+
+    snapshot_location = get_snapshot_location(database)
+    puts "Removing snapshot #{snapshot_location}"
+    run_cmd("lvremove -f #{snapshot_location}")
+  end
+end
+
+def online_backup(database)
+  case database
+  when 'pulp'
+    FileUtils.cd '/var/lib/pulp' do
+      puts "Backing up Pulp data... "
+      matching = false
+      until matching
+        checksum1 = `find . -printf '%T@\n' | md5sum`
+        create_pulp_data_tar
+        checksum2 = `find . -printf '%T@\n' | md5sum`
+        matching = (checksum1 == checksum2)
+      end
+    end
+    puts "Done."
+  when 'pgsql'
+    puts "Backing up postgres online schema... "
+    run_cmd("runuser - postgres -c 'pg_dumpall -g > #{File.join(@dir, 'pg_globals.dump')}'")
+    run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{File.join(@dir, 'foreman.dump')}'")
+    run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{File.join(@dir, 'candlepin.dump')}'")
+    puts "Done."
+  when 'mongodb'
+    puts "Backing up mongo online schema... "
+    run_cmd("mongodump --host localhost --out #{File.join(@dir, 'mongo_dump')}")
+    puts "Done."
+  end
+end
+
+def offline_backup(database, dir_path = nil)
+  case database
+  when 'pulp'
+    dir_path ||= '/var/lib/pulp'
+    FileUtils.cd dir_path do
+      puts "Backing up Pulp data... "
+      create_pulp_data_tar
+      puts "Done."
+    end
+  when 'mongodb'
+    dir_path ||= '/var/lib/mongodb'
+    FileUtils.cd dir_path do
+      puts "Backing up mongo db... "
+      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'mongo_data.tar')} --listed-incremental=#{File.join(@dir, '.mongo.snar')} --exclude=mongod.lock --transform 's,^,var/lib/mongodb/,S' -S *")
+      puts "Done."
+    end
+  when 'pgsql'
+    dir_path ||= '/var/lib/pgsql/data'
+    FileUtils.cd dir_path do
+      puts "Backing up postgres db..."
+      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pgsql_data.tar')} --listed-incremental=#{File.join(@dir, '.postgres.snar')} --transform 's,^,var/lib/pgsql/data/,S' -S *")
+      puts "Done."
+    end
+  end
 end
 
 def generate_metadata
+  puts "Generating metadata ... "
   os_version = `cat /etc/redhat-release`.chomp
   plugin_list = `foreman-rake plugin:list | grep 'Foreman plugin: '`.lines
   plugins = []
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -54,49 +95,81 @@ def run_cmd(command, exit_codes=[0])
...
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def snapshot_backup
+  snapsize = @options[:snapshot_size] || "2G"
+  mountdir = "#{@options[:snapshot_mount]}/rootsnapshot" || "/mnt/snap/rootsnapshot"
+  FileUtils.mkdir_p mountdir
+  run_cmd("katello-service stop #{EXCLUDED}")
+  run_cmd("lvcreate -s -nrootsnapshot -L#{snapsize} /dev/#{@options[:snapshot]}/root")
+  run_cmd("mount /dev/#{@options[:snapshot]}/rootsnapshot #{mountdir} -txfs -onouuid,ro")
+  run_cmd("katello-service start #{EXCLUDED}")
```
sean797: Small thing but I think running `--exclude foo,bar` is kind of unnecessary here

katello/katello-backup
```diff
@@ -54,49 +95,81 @@ def run_cmd(command, exit_codes=[0])
...
-
-  puts "Backing up mongo online schema... "
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def snapshot_backup
+  snapsize = @options[:snapshot_size] || "2G"
+  mountdir = "#{@options[:snapshot_mount]}/rootsnapshot" || "/mnt/snap/rootsnapshot"
+  FileUtils.mkdir_p mountdir
+  run_cmd("katello-service stop #{EXCLUDED}")
+  run_cmd("lvcreate -s -nrootsnapshot -L#{snapsize} /dev/#{@options[:snapshot]}/root")
```
sean797: What if my LV is named something other than root?
What if I have separate partitions for other mount points?

katello/katello-backup
```diff
@@ -110,6 +183,11 @@ if @dir.nil?
   puts "**** ERROR: Please specify an export directory ****"
   puts optparse
   exit(-1)
+elsif
+  @options[:snapshot] && (@options[:online] || @options[:incremental])
+  puts "**** ERROR: Snapshot backups can only be done offline ****"
```
sean797: `ERROR: Snapshot backups cannot be done online or incrementally` ?

Also, why can't we do them online?

sean797 (changes requested):
TBH I don't think this fits in our scripts. 

If you want to use LVM snapshots its only 1 cmd per mount point, only the user knows the mount points and is going to be wildly different on each install.

IMO its kind of pointless to essentially use katello-backup to wrap `lvcreate -s..... `

Also what about katello-restore?

katello/katello-backup
cfouant: I'm pretty new to logical volumes, so I'm sure my knowledge base is missing some information. I was under the impression that to take a snapshot of root, which is where all the files we look to backup are located, it would be in `/dev/<volume group name>/root`. Is this incorrect?

katello/katello-backup
cfouant: Respectfully disagree on this point. Since we aren't stopping the excluded services, we shouldn't attempt to start them, if only for the few seconds of time it may save. If the excluded services should grow, this keeps it more manageable and obvious as well.

katello/katello-backup
cfouant: I was under the impression that the snapshot must be created while services were shut off. If that isn't the case, I'm happy to add an online solution. I'll work on updating this one now.

katello/katello-backup
```diff
@@ -54,49 +95,85 @@ def run_cmd(command, exit_codes=[0])
...
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
-  puts "Done."
+def snapshot_backup
+  snapsize = @options[:snapshot_size] || "2G"
+  mountdir = "#{@options[:snapshot_mount]}/rootsnapshot" || "/mnt/snap/rootsnapshot"
+  FileUtils.mkdir_p mountdir
+  run_cmd("katello-service stop #{EXCLUDED}") unless @options[:online]
+  run_cmd("lvcreate -s -nrootsnapshot -L#{snapsize} /dev/#{@options[:snapshot]}/root")
+  run_cmd("mount /dev/#{@options[:snapshot]}/rootsnapshot #{mountdir} -txfs -onouuid,ro")
```
evgeni: `-txfs` implies the user has XFS, which they don't have to. I would just omit passing `-t` and let `mount` handle it automatically.

cfouant:
@thomasmckay @bbuckingham - are there any members of your team available to review this? I can help on the setup if needed, but it would be great if there were any who already understood lvm filesystem so if I have errors in my knowledge base, I'm not passing on misinformation!

katello/katello-backup
```diff
@@ -54,49 +95,85 @@ def run_cmd(command, exit_codes=[0])
...
-  run_cmd("runuser - postgres -c 'pg_dumpall -g > #{@dir}/pg_globals.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
-  puts "Done."
+def snapshot_backup
+  snapsize = @options[:snapshot_size] || "2G"
+  mountdir = "#{@options[:snapshot_mount]}/rootsnapshot" || "/mnt/snap/rootsnapshot"
+  FileUtils.mkdir_p mountdir
+  run_cmd("katello-service stop #{EXCLUDED}") unless @options[:online]
+  run_cmd("lvcreate -s -nrootsnapshot -L#{snapsize} /dev/#{@options[:snapshot]}/root")
```
evgeni: * this will fail if there is already a snapshot called `rootsnapshot` (which the user could have created them-self for some reason)
* this implies that all data is on the same LV with the name `root`. a prod deployment usually will have `/var`, `/var/lib/pgsql/`, `/var/lib/mongodb` and `/var/lib/pulp` on separate LVs, with varying names.

evgeni: Also we should probably check if the VG actually has enough free space, even before stopping the services

evgeni (changes requested):

evgeni:
I think the snapshotting logic should be more like this (pseudo code, but I think you get the idea):

```ruby
snaps = []
DATABASES.each do |db|
  lv = find_lv_for_mountpoint(db.mountpoint)
  snap = lv.create_snapshot
  snaps << snap
end
backup
snaps.each do |snap|
  snap.umount
  snap.remove
end
```

katello/katello-backup
```diff
@@ -34,6 +61,20 @@ optparse = OptionParser.new do |opts|
     @options[:logical_backup] = logical
   end
 
+  opts.on("--snapshot [VOLUME_GROUP_LOCATION]", String, "Use LVM snapshots to perform an offline backup") do |volume_group|
```
evgeni: VG location seems to imply a path, so I tried `--snapshot /dev/VolGroup00` and it failed. I think it should say "VG name" here? And we could check for the name not having any `/` in it to be sure the user gave a name, not a path.

evgeni:
That failed in my test VM with

```
Backing up mongo db... 
tar: var/lib/mongodb/journal/prealloc.0: File shrank by 1071837184 bytes; padding with zeros
tar: var/lib/mongodb/journal/prealloc.1: Read error at byte 0, while reading 2048 bytes: Input/output error
tar: var/lib/mongodb/journal/prealloc.2: Read error at byte 0, while reading 4608 bytes: Input/output error
tar: Exiting with failure status due to previous errors
Failed 'tar --selinux --create --file=/var/tmp/katello-backup/katello-backup-20170509174658/mongo_data.tar --listed-incremental=/var/tmp/katello-backup/katello-backup-20170509174658/.mongo.snar --exclude=mongod.lock var/lib/mongodb/'
```
I think this happened because the snapshot I created was too small, so we might want to enforce a certain minimum size (esp as I was writing the backup to the same LV as the snapshot was, which seems like a terrible idea).

evgeni:
Another thing: I would have thought that `--snapshot` and `--online-backup` are mutually exclusive as you want to have a snap of the on-disk data, which does not seem possible while the services are online?

evgeni (changes requested):

katello/katello-backup
```diff
@@ -54,49 +95,85 @@ def run_cmd(command, exit_codes=[0])
...
 
-def backup_db_online
-  puts "Backing up postgres online schema... "
-  run_cmd("runuser - postgres -c 'pg_dumpall -g > #{@dir}/pg_globals.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
-  puts "Done."
+def snapshot_backup
+  snapsize = @options[:snapshot_size] || "2G"
+  mountdir = "#{@options[:snapshot_mount]}/rootsnapshot" || "/mnt/snap/rootsnapshot"
```
evgeni: For some reason it created `/rootsnapshot/` for me, and not `/mnt/snap/rootsnapshot/`.

evgeni:
Some brain dump from an offline discussion with @cfouant in the office:
* snapshot and online should be exclusive, as otherwise we would end up with a live online backup of pgsql and mongo (as we use export commands, talking to a running daemon, which reads the non-snapshotted data) and an old backup of pulp (as it uses the data from the snapshot). this pretty sure makes the whole backup inconsistent and error prone
* we think that katello-restore should not need any adjustments, as the prereq for a restore is a default install of katello, thus all mountpoints/LVs/etc should be already preset and are up to the user (and thus its not any change to the old behavior)
* as we can't possibly know how the user has their LVs set up, we should probably just ask the user for a list of mountpoints that are on separate LVs and use that list for the snapshots.

@cfouant - did I miss anything?
@sean797 - does the above make sense to you?

sean797:
@cfouant @evgeni most of the above makes sense though it does lead me to the same conclusion as before.

This is how I see it:
- I don't think we should try to become an all singing, all dancing backup script with loads of features/options. There are plenty of other tools dedicated to this that will do a better job.
   - If we add LVM snapshots, should we add rsync, Relax-and-Recover, backuping up to ISOs, where does this stop ?
- If we need to ask the user for every option we pass to `lvcreate -s` then what benefit does this actually have? What's the point in a wrapper script if it doesn't do anything cleaver?
- There are many possible options valid for `lvcreate -s`. What if I want to use ` --thinpool`, `--extents` or others, do we accept values for all these options? Where does this end?
- An inexperienced sysadmin with knowledge of their LVs set up could script backups using LVM snapshots in 15 minutes and learn something along the way.
- Adds complexity to this script

Sorry if I sound super negative about this, I just see katello-backup as a simple, ~~quick~~ & cheap backup.

I see this as a DOC bug, people need to know what directories they need to backup if they would like to use other tools, `katello-backup list` to list directories requiring backup would suffice IMO.

cfouant:
@sean797 - I definitely see where you're coming from, and agree a lot in limiting features; however, snapshots are a wildly anticipated feature that's definitely going to happen. I was able to limit snapshots to just LVM, which I mention because there were requests for 3 different snapshot options!

katello/katello-backup
```diff
@@ -3,24 +3,52 @@
...
+  '/opt/puppetlabs/puppet/ssl',
+  '/var/lib/puppet/foreman_cache_data',
+  '/var/lib/puppet/ssl'
+]
 
 @options = {}
 @dir = nil
-
-EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
+@databases = DATABASES
```
evgeni: this should be `@databases = DATABASES.dup`, otherwise:

```ruby
irb(main):001:0> FOO = [1,2,3]
=> [1, 2, 3]
irb(main):002:0> @foo = FOO
=> [1, 2, 3]
irb(main):003:0> @foo.delete 1
=> 1
irb(main):004:0> @foo
=> [2, 3]
irb(main):005:0> FOO
=> [2, 3]
```

katello/katello-backup
```diff
@@ -35,6 +63,22 @@ optparse = OptionParser.new do |opts|
...
 
+  opts.on("--snapshot", "Locations of the logical volumes of #{DATABASES.join(',')}") do |snapshot|
+    @options[:snapshot] = snapshot
+  end
+
+  opts.on("--snapshot-size SNAPSHOT_BLOCK_DEVICE_SIZE", String, "Override default block size (5G)") do |size|
+    @options[:snapshot_size] = size
+  end
+
+  opts.on("--snapshot-filetype FILE_SYSTEM_TYPE", String, "Override default filesystem type (xfs)") do |type|
```
evgeni: I don't this we need the fstype, just ommit it and let the kernel handle it for you

katello/katello-backup
```diff
@@ -63,59 +107,125 @@ def create_directories(directory)
...
-    matching = (checksum1 == checksum2)
+def create_and_mount_snapshots
+  @databases.each do |database|
+    puts "Creating and mounting #{database} snapshot"
+    lv_location = get_lv_location(database)
+    run_cmd("lvcreate -n#{database}-snap -L#{@snapsize} -s #{lv_location}")
+    FileUtils.mkdir_p "#{File.join(@mountdir, database)}"
+    snapshot_location = get_snapshot_location(database)
+    puts snapshot_location
+    run_cmd("mount /dev/rhel/#{database}-snap #{File.join(@mountdir, database)} -t#{@filetype}")
```
evgeni: I don't this we need the fstype, just ommit it and let the kernel handle it for you

evgeni: please don't hardcode the VG as `rhel`, users will have various names here. `#{snapshot_location}` should be exactly what you need.

katello/katello-backup
```diff
@@ -63,59 +107,125 @@ def create_directories(directory)
...
+  target = "/var/lib/"
+  case database
+  when 'pulp'
+    target << 'pulp'
+  when 'pgsql'
+    target << 'pgsql/data'
+  when 'mongod'
+    target << 'mongodb'
+  end
+  `findmnt -n --target #{target}`.split(' ')[1]
```
evgeni: `findmnt -n -o SOURCE --target #{target}`

`-o SOURCE` outputs only the source column, so you don't have to `split()`

katello/katello-backup
```diff
@@ -63,59 +107,125 @@ def create_directories(directory)
...
+end
 
-  puts "Backing up mongo db... "
-  run_cmd('tar --selinux --create --file=mongo_data.tar --listed-incremental=.mongo.snar --exclude=mongod.lock /var/lib/mongodb/')
-  puts "Done."
+def get_lv_location(database)
+  target = "/var/lib/"
+  case database
+  when 'pulp'
+    target << 'pulp'
```
evgeni: `File.join`? :)

evgeni (changes requested):

katello/katello-backup
cfouant: However, if the file type is xfs, there needs to be the -onouuid,ro tag added. Without it, mounting fails.

katello/katello-backup
cfouant: That was an accident/old code, thanks for seeing it!

katello/katello-backup
cfouant: I do love that new trick you taught me :) will change

katello/katello-backup
cfouant: AWESOME, thanks

katello/katello-backup
cfouant: Noted, will change

katello/katello-backup
evgeni: Mhh. Right.
I'd even argue that `-o ro` should *always* be added.

katello/katello-backup
```diff
@@ -63,59 +100,132 @@ def create_directories(directory)
...
 
-  puts "Backing up mongo online schema... "
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def create_and_mount_snapshots
+  @databases.each do |database|
+    puts "Creating and mounting #{database} snapshot"
+    run_cmd("lvcreate -n#{database}-snap -L#{get_snap_size(database)} -s #{get_lv_location(database)}")
+    FileUtils.mkdir_p "#{File.join(@mountdir, database)}"
+    run_cmd("mount #{get_snapshot_location(database)} #{File.join(@mountdir, database)} -onouuid,ro")
```
evgeni: Heh, I fear now I talked you into something that won't work either :(
`nouuid` is specific for xfs, so we can't pass it unconditionally. `ro` is OK for all fstypes, though.

katello/katello-backup
```diff
@@ -63,59 +100,132 @@ def create_directories(directory)
...
+  case database
+  when 'pulp'
+    target = 'pulp'
+  when 'pgsql'
+    target = 'pgsql/data'
+  when 'mongodb'
+    target = 'mongodb'
   end
-  puts "Done."
+  run_cmd("du -sh #{File.join('/var/lib', target)}").split.first
```
evgeni: that creates snapshots the size of the current disk usage. that means if I have 600G pulp, I need 600G free space for the snapshot. ain't nobody has so much space left :(

katello/katello-backup
```diff
@@ -63,59 +100,132 @@ def create_directories(directory)
...
+  create_and_mount_snapshots
+  run_cmd("katello-service start #{EXCLUDED}")
+  backup_and_destroy_snapshots
+end
 
-  puts "Backing up mongo online schema... "
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def create_and_mount_snapshots
+  @databases.each do |database|
```
evgeni: this creates one snapshot per database, in the case where multiple dbs are on the same LV, we thus create multiple snaps of the same data. not too bad (we do that in the "split everything" case too) but one could safe a bit of space here. would make the code more complicated, though

evgeni (changes requested):

cfouant:
[test]

katello/katello-backup
```diff
@@ -63,59 +104,119 @@ def create_directories(directory)
...
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
-  puts "Done."
-
-  puts "Backing up mongo online schema... "
-  run_cmd('mongodump --host localhost --out mongo_dump')
-  puts "Done."
+def snapshot_backup
+  @mountdir = @options[:snapshot_mount] || "/var/snap"
+  @snapsize = @options[:snapshot_size] || "2G"
```
evgeni, line 109: above you say the default is 5G, what is correct? ;)

katello/katello-backup
```diff
@@ -63,59 +104,119 @@ def create_directories(directory)
...
-    matching = (checksum1 == checksum2)
+def create_and_mount_snapshots
+  @databases.each do |database|
+    puts "Creating and mounting #{database} snapshot"
+    lv_info = get_lv_info(database)
+    run_cmd("lvcreate -n#{database}-snap -L#{@snapsize} -s #{lv_info[0]}")
+    FileUtils.mkdir_p "#{File.join(@mountdir, database)}"
+    snapshot_location = run_cmd("lvs --noheadings -o lv_path -S lv_name=#{database}-snap").strip
+    options = lv_info[1] == 'xfs' ? "-onouuid,ro" : "-oro"
+    run_cmd("mount #{get_snapshot_location(database)} #{File.join(@mountdir, database)} #{options}")
```
evgeni: This errors out with:
```
Creating and mounting pulp snapshot
katello-backup:125:in `block in create_and_mount_snapshots': undefined method `get_snapshot_location' for main:Object (NoMethodError)
	from katello-backup:118:in `each'
	from katello-backup:118:in `create_and_mount_snapshots'
	from katello-backup:112:in `snapshot_backup'
	from katello-backup:271:in `<main>'
```

evgeni (changes requested):

katello/katello-backup
cfouant: @evgeni - ah, i fixed it, that method didn't make it in to the script i pushed. Please try again, working now!

evgeni:
Current version fails for me with:

```
Backing up mongo db... 
tar: journal/j._0: file changed as we read it
Failed 'tar --selinux --create --file=/dev/shm/backup/katello-backup-20170613064103/mongo_data.tar --listed-incremental=/dev/shm/backup/katello-backup-20170613064103/.mongo.snar --exclude=mongod.lock --transform 's,^,var/lib/mongodb/,S' -S *'
```

This is due to https://github.com/Katello/katello-packaging/pull/414#discussion_r121596736

katello/katello-backup
```diff
@@ -63,59 +104,122 @@ def create_directories(directory)
...
-  run_cmd('tar --selinux --create --file=mongo_data.tar --listed-incremental=.mongo.snar --exclude=mongod.lock /var/lib/mongodb/')
-  puts "Done."
+def get_lv_info(database)
+  target = ""
+  case database
+  when 'pulp'
+    target = 'pulp'
+  when 'pgsql'
+    target = 'pgsql/data'
+  when 'mongod'
```
evgeni: the database name in the array is `mongodb`, not `mongod`

katello/katello-backup
```diff
@@ -63,59 +104,122 @@ def create_directories(directory)
...
-  create_pulp_data_tar
-  puts "Done."
+def get_base_directory(database)
+  target = ""
+  case database
+  when 'pulp'
+    target = '0005_puppet_module_name_change.txt'
+  when 'pgsql'
+    target = 'postgresql.conf'
+  when 'mongod'
```
evgeni: the database name in the array is `mongodb`, not `mongod`

katello/katello-backup
```diff
@@ -63,59 +104,122 @@ def create_directories(directory)
...
+    target = 'mongod.lock'
+  end
+  Find.find(File.join(@mountdir, database)) do |path|
```
evgeni: this loop will make the function return `nil` when when the target is not the last item in the list. I would suggest using something like this:

```ruby
  result = nil
  Find.find(File.join(@mountdir, database)) do |path|
    result = File.dirname(path) if File.basename(path) == target
  end
  result
```

evgeni (changes requested):

katello/katello-backup
cfouant: Updated

katello/katello-backup
cfouant: ah, great catch, updating now!

evgeni:
ACK on this PR from me. @ehelms @johnpmitsch @sean797 @chris1984 what do you all think?

However, I think we should prioritize the following followups:
* add tests for backup/restore (via bats in forklift?)
* add error handling to at least restore katello services after the backup failed, as currently a failed backup will also leave the services down :/

evgeni (approved):

cfouant:
@evgeni - there's a separate PR for error handling, if you want to review that! https://github.com/Katello/katello-packaging/pull/451

@johnpmitsch has started on a test suite, the project is in the beginning stages, but he is creating a PR when it's come a little further

johnpmitsch:
@evgeni I started working on some ansible playbooks that will perform different backup/restore scenarios on a katello instance, I have there [here](https://github.com/johnpmitsch/katello-backup-restore-test) for now, once they are reliable we can talk with @ehelms about working them into a CI/CD system. I would rather test on the PR than in bats, but we can discuss this more. For now, they are something to run locally to test PRs.

johnpmitsch (approved):
I didn't test the functionality (I'm assuming @evgeni has), but the code looks fine to me

cfouant:
[test]

sean797:
By the way I'm not very happy with this being merged 20 minutes after the first ACK, especially considering the comments I made in https://github.com/Katello/katello-packaging/pull/414#issuecomment-300616907, I left my 'requesting changes' flag set and the only reply being:
> agree a lot in limiting features; however, snapshots are a wildly anticipated feature that's definitely going to happen.

Saying something is "definitely going to happen." is not valid. I also don't seen any evidence of any users requesting this, my search on foreman-users yields nothing.

I was actually planning on taking another look at this when @cfouant was happy with it to check all my previous points were still valid and actually get a response other than its "definitely going to happen".

But whatever. Just in future we should try to counter people views with facts and/or opinions & allow more than 20 minutes before being merged, especially when someone has expressed another opinion or view.

cfouant:
@sean797 - I did request feedback from users on a satellite6 email thread, and got few responses. I am not sure if you're on that list or not. There are 4 open customer cases regarding using snapshots in BZ. I also responded to your comment on the 15th, and this merged a month after that, with no feedback from you in that interim.

PR #419, opened 2017-04-10
Create katello-agent restart file only when needed.

zjhuntin:
Wanted someone to look at this before creating a redmine about it, but will update with redmine number if/once it's been acknowledged as being needed.

### DIFF
--- a/katello-agent/katello-agent.spec
+++ b/katello-agent/katello-agent.spec
@@ -88,13 +88,21 @@ cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}
 rm -rf %{buildroot}
 
 %post
-chkconfig goferd on
 %if 0%{?fedora} > 18 || 0%{?rhel} > 6
-  /bin/systemctl start goferd > /dev/null 2>&1 || :
+  systemctl enable goferd
+  if [systemctl status goferd] then
+    touch /tmp/katello-agent-restart
+  else
+    /bin/systemctl start goferd > /dev/null 2>&1 || :
+  fi
 %else
-  /sbin/service goferd start > /dev/null 2>&1 || :
+  /sbin/chkconfig goferd on
+  if [service goferd status] then
+    touch /tmp/katello-agent-restart
+  else
+    /sbin/service goferd start > /dev/null 2>&1 || :
+  fi
 %endif
-touch /tmp/katello-agent-restart
 exit 0
 
 %posttrans


katello-agent/katello-agent.spec
```diff
@@ -88,13 +88,21 @@ cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}
 rm -rf %{buildroot}
 
 %post
-chkconfig goferd on
 %if 0%{?fedora} > 18 || 0%{?rhel} > 6
-  /bin/systemctl start goferd > /dev/null 2>&1 || :
+  chkconfig goferd on
```
stbenjam: Nitpick, If we're moving this into the systemd distro block, we should move to using `systemctl enable goferd`

katello-agent/katello-agent.spec
```diff
@@ -88,13 +88,21 @@ cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}
...
+  if systemctl status goferd | grep 'active (running)'; then
```
stbenjam: You can probably just rely on the systemctl exit code instead of a grep.

```sh
sh-4.3$ if systemctl status firewalld 2>&1 > /dev/null; then echo "hello"; fi
hello
sh-4.3$ sudo systemctl stop firewalld
sh-4.3$ if systemctl status firewalld 2>&1 > /dev/null; then echo "hello"; fi
sh-4.3$
```

stbenjam:
But looks good to me otherwise

zjhuntin:
@stbenjam Thanks for the comments! Updated.

sean797:
Why bother using the restart file at all? Is there any benefits to using it?
Just doing `systemctl restart goferd` will restart or start if it's not running.

Also ping @zjhuntin tests failed 😭

stbenjam:
> Why bother using the restart file at all? Is there any benefits to using it?
Just doing systemctl restart goferd will restart or start if it's not running.

~~Yes, if you have 50,000 servers, it's not that simple.  We need to ensure the daemon is restarted when updated the RPM.~~

Oh nevermind, I misunderstood the question.  The reason for the restart file is you can update the katello-agent RPM through katello-agent itself, so the restart needs to be delayed until the RPM transaction is finished.  With the restart file, the agent waits until the pending jobs are finished before restart.

katello-agent/katello-agent.spec
```diff
@@ -88,13 +88,21 @@ cp extra/katello-agent-send.cron %{buildroot}%{_sysconfdir}/cron.d/%{name}
 rm -rf %{buildroot}
 
 %post
-chkconfig goferd on
 %if 0%{?fedora} > 18 || 0%{?rhel} > 6
-  /bin/systemctl start goferd > /dev/null 2>&1 || :
+  systemctl enable goferd
+  if [systemctl status goferd] then
```
stbenjam, line 93: Brackets aren't needed, and this actually won't work, [ and [[ expect a condition.  I'd use something closer to my suggestion, especially with redirecting output to /dev/null.

sean797:
@stbenjam that doesn't make sense to me but I think I would the answer in case anyone else was wondering :-)

>  At the moment when katello-agent is updated from the katello web-ui it fails because in katello-agents packaging it restarts goferd. Goferd is running this process so it fails.

http://projects.theforeman.org/issues/18187

zjhuntin:
This actually may not be needed. I'm going to close it for now.

PR #425, opened 2017-04-13
Fixes #19263 - dont try and rm a mounted directory

sean797:
We could do the same for the other arrays but I doubt anyone has them mounted separately.

I've also removed `/var/cache/foo` directories as we already remove `/var/cache/` so they are kind of pointless.

### DIFF
--- a/katello/katello-remove
+++ b/katello/katello-remove
@@ -107,7 +107,6 @@ CERT_FILES=[
 ]
 
 CONTENT=[
-  '/var/cache/pulp',
   '/usr/share/foreman-proxy',
   '/usr/share/foreman-installer-katello',
   '/var/www/html/pub/katello-server-ca.crt',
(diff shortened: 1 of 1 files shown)


katello/katello-remove
```diff
@@ -129,13 +128,19 @@ CONTENT=[
   '/usr/share/katello-installer-base',
   '/usr/share/qpid',
   '/usr/share/qpid-tools',
-  '/var/cache/candlepin',
-  '/var/cache/foreman-proxy',
```
jlsherrill, line 133: whats the reason for removing the /var/cache/ directories?

sean797: `/var/cache/` is listed below so they were completely pointless

https://github.com/Katello/katello-packaging/blob/master/katello/katello-remove#L134

jlsherrill: oh right, i should have read the description a bit closer  jlsherrill--  👍

jlsherrill:
ACK thanks @sean797 !

PR #427, opened 2017-04-17
Fixes #19281 - Unable to run online backup via cron

ahumbe:
Fix is for the error while backing up foreman and candlepin database via katello-backup script via cron:

Backing up postgres db...
sh: runuser: command not found
sh: runuser: command not found
Done.

When we execute any cron job the default PATH is /usr/bin:/bin since the runuser command is present in /usr/sbin/runuser which cron can not access and execute, so backup script fails with above error. Running katello-backup script manually as a root user work without any error, issue is only when we run it via cron.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -56,9 +56,9 @@ end
 
 def backup_db_online
   puts "Backing up postgres online schema... "
-  run_cmd("runuser - postgres -c 'pg_dumpall -g > #{@dir}/pg_globals.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc foreman > #{@dir}/foreman.dump'")
-  run_cmd("runuser - postgres -c 'pg_dump -Fc candlepin > #{@dir}/candlepin.dump'")
+  run_cmd("/usr/sbin/runuser - postgres -c 'pg_dumpall -g > #{@dir}/pg_globals.dump'")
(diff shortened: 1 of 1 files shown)


jlsherrill:
[test]

ahumbe:
Can some one review this PR ? We have 3 customers reported this issue till now.

evgeni:
IMHO we should not hard code the path to the utility, but the cron environment should be set up correctly.

ahumbe:
In general linux admin practice most of the commands which can be executed from bash as a root are executable via cron also, so not all users are aware that they need to set the PATH specifically for the katello-backup script.

evgeni:
Right, and that's why I would have expected `katello-backup` to be in sbin itself. It's useless for normal users as those can't access most of the directories that have to be backed up anyways.

jlsherrill:
Should we just change this issue then to put all of those scripts in sbin instead?

evgeni:
:+1:

cfouant:
ACK

cfouant:
@jlsherrill can we merge?

johnpmitsch:
@jlsherrill @evgeni based on the conversation above, it sounds like we want to update the spec file to place the scripts in /usr/sbin. Should we open up a new issue or change the existing one and close this PR if we are going that route?

evgeni:
@johnpmitsch I'd prefer that, yeah

johnpmitsch:
Closing based on above discussion to add these scripts to sbin

PR #428, opened 2017-04-19
Fixes #17060 - Addition to katello-debug.sh

ahumbe:
Added below details 

- Collect below files from /etc/pulp/server/plugins.conf.d/ which include the proxy configuration details
docker_importer.json, iso_importer.json, ostree_distributor.json,  ostree_importer.json, puppet_importer.json, yum_importer.json

- Collect katello-service status command output which is helpful while debugging issues

Removed : 
- Removed entry for /etc/pulp/server/plugins.conf.d/nodes/distributor/* since we are not using nodes feature and the directory does not exist on satellite.


jlsherrill:
[test]

jlsherrill:
ACK thanks @ahumbe !

PR #429, opened 2017-04-20
call tito with --builder mock

evgeni:
builder.MockBuilder is wrong, it is either tito.builder.MockBuilder (thefull module name) or mock which is defined as an alias.


jlsherrill:
[test]

ehelms (approved):

PR #430, opened 2017-04-21
python-urllib3: use https links in spec file

evgeni:
otherwise git-annex chokes on the redirect
while at it, also fix up readthedocs url


jlsherrill:
[test]

ehelms:
This should actually be pulled from the repo now I believe. The version in EPEL is greater than the version we have here now.

ehelms:
I think this can be replaced by https://github.com/Katello/katello-packaging/pull/471

PR #359, opened 2017-01-23
#358 allow globbing to catch the *.snar files during incremental backups

ericlake:

PR #436, opened 2017-05-03
fixes #19446 - fixes incremental backup performing full backup

cfouant:


zjhuntin:
I don't seem to be able to get this working properly...

I ran katello-backup /tmp/backup

Which seemed to work fine. Then I created a new repository and synced it.

After that I ran katello-backup /tmp/backup --incremental /path/to/last/backup

Which resulted in:

https://gist.github.com/zjhuntin/0e8a816b4cf7269ee2daa05fadec3746

Am I performing the recreation steps incorrectly?

katello/katello-backup
```diff
@@ -160,7 +160,7 @@ else
   run_cmd("tar --selinux --create --gzip --file=config_files.tar.gz --listed-incremental=.config.snar #{CONFIGS.join(' ')} 2>/dev/null", [0,2])
   puts "Done."
 
-  `shopt -s dotglob ; cp #{@options[:incremental]}/*.snar .` if @options[:incremental]
+  `shopt -s dotglob ; cp #{@options[:incremental]}/.*.snar .` if @options[:incremental]
```
pondrejk: This way the wildcard won't get expanded and the copy will fail with `cp: cannot stat ‘.../.*.snar’ : No such file or directory`
I suggest using `FileUtils.cp`, fileutils are already imported in the script

zjhuntin:
This seemed to work for me:

I made a fresh katello install then ran the katello-backup on it, then ran an incremental backup after adding a repository.

https://paste.fedoraproject.org/paste/0fAItX9zDEral5h8uNgDvF5M1UNdIGYhyRLivL9gydE=

If more information is needed, I'd be happy to provide it.

stbenjam:
Unless I'm missing something, the output of @zjhuntin's backup doesn't look like incremental, the incremental dir doesn't have any snar's and sizes all look like full backups.

zjhuntin:
@stbenjam I think that I did not run that first one with ls -la instead just ls -l.

Here is an updated one 

https://paste.fedoraproject.org/paste/dGPxj77~QcU0RLYZQvwYgF5M1UNdIGYhyRLivL9gydE=

cfouant:
@evgeni - if you could test this when you have a second, it would be much appreciated!

evgeni:
Heh.

So actually this was already fixed in https://github.com/Katello/katello-packaging/pull/359 by @ericlake :D
http://projects.theforeman.org/issues/19446 aka https://bugzilla.redhat.com/show_bug.cgi?id=1445989 is only about the downstream version which did not get #359 

Now, said that, I like your version a tad more, as it does not use `shopt -s dotglob` and actually even better, removes yet another line of crud shell code from a Ruby tool :+1: 
Plus it enables incremental backups for /etc, not that it would be really needed, but that was an oversight in the original code.

FTR, test was done with:
```
./katello-backup --skip-pulp-content /mnt/backup
./katello-backup --skip-pulp-content --incremental /mnt/backup/katello-backup-20170510210744/ /mnt/backup
```

The first folder is about 800M big, the second only 100.
The result with the original code from master is identical.

cfouant:
So, that PR #359 actually didn't end up catching the files - still needed the "." character before the wildcard. I also prefer switching the code over to ruby and I'm glad I found it!

evgeni (approved):

evgeni:
ACK

cfouant:
@ehelms @jlsherrill can I have merge pls?

mccun934:
merged, as @ehelms and @jlsherrill are out

PR #441, opened 2017-05-09
Fixes #17036 - Adding katello-backup and restore for a proxy

johnpmitsch:
See [comment below](https://github.com/Katello/katello-packaging/pull/441#issuecomment-305794770) for review instructions

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -5,38 +5,19 @@ require 'fileutils'
 require 'date'
 require 'yaml'
 require 'find'
+require 'json'
 
 EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
 DATABASES = ['pulp', 'pgsql', 'mongodb']
-CONFIGS = [
-  '/etc/candlepin',
-  '/etc/foreman',
-  '/etc/foreman-proxy',
-  '/etc/hammer',
-  '/etc/httpd',
-  '/etc/foreman-installer',
-  '/etc/pki/katello',
-  '/etc/pki/katello-certs-tools',
-  '/etc/pki/pulp',
-  '/etc/pulp',
-  '/etc/puppet',
-  '/etc/puppetlabs',
-  '/etc/qpid',
-  '/etc/qpid-dispatch',
-  '/etc/sysconfig/tomcat*',
-  '/etc/tomcat*',
-  '/root/ssl-build',
-  '/var/lib/candlepin',
-  '/var/www/html/pub',
-  '/opt/puppetlabs/puppet/cache/foreman_cache_data',
-  '/opt/puppetlabs/puppet/ssl',
-  '/var/lib/puppet/foreman_cache_data',
-  '/var/lib/puppet/ssl'
-]
 
 @options = {}
 @dir = nil
 @databases = DATABASES.dup
+@is_foreman_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
+
+# keep as variables for easy backporting
+FOREMAN_PROXY_CONTENT = "foreman-proxy-content"
+FOREMAN_PROXY = "foreman proxy"
 
 optparse = OptionParser.new do |opts|
   opts.banner = "Usage: katello-backup /path/to/dir [options]\n eg: $ katello-backup /tmp/katello-backup"
@@ -75,6 +56,11 @@ optparse = OptionParser.new do |opts|
     @options[:snapshot_size] = size
   end
 
+  opts.on("--features FEATURES", Array, "#{FOREMAN_PROXY.capitalize} features to include in the backup, please specify a list with commas. " \
+          "Valid features are tftp, dns, dhcp, openscap, and all.") do |features|
+    @options[:features] = features.map { |f| f.to_s.downcase }
+  end
+
   opts.parse!
 
   if ARGV.length == 0
@@ -85,6 +71,67 @@ optparse = OptionParser.new do |opts|
   end
 
   @dir = ARGV[0].dup
+end
+
+@databases.delete 'pgsql' if @is_foreman_proxy_content
+
+def specified_features
+  @options[:features] || []
+end
+
+def feature_included?(feature)
+  specified_features.include?(feature) || specified_features.include?("all")
+end
+
+def configure_configs
+  base_configs = [
+    "/etc/foreman-proxy",
+    "/etc/httpd",
+    "/etc/foreman-installer",
+    "/etc/pki/katello",
+    "/etc/pki/katello-certs-tools",
+    "/etc/pki/pulp",
+    "/etc/pulp",
+    "/etc/puppet",
+    "/etc/qpid",
+    "/etc/qpid-dispatch",
+    "/root/ssl-build",
+    "/var/www/html/pub",
+    "/etc/squid",
+    "/etc/puppetlabs",
+    '/opt/puppetlabs/puppet/cache/foreman_cache_data',
+    '/opt/puppetlabs/puppet/ssl/',
+    '/var/lib/puppet/foreman_cache_data',
+    '/var/lib/puppet/ssl'
+  ]
+
+  katello_configs = [
+    "/etc/candlepin",
+    "/etc/foreman",
+    "/etc/hammer",
+    "/etc/sysconfig/tomcat*",
+    "/etc/tomcat*",
+    "/var/lib/candlepin",
+  ]
+
+  if @is_foreman_proxy_content
+    fpc_certs_tar = `cat /etc/foreman-installer/scenarios.d/#{FOREMAN_PROXY_CONTENT}-answers.yaml | grep certs_tar | awk '{print $2}'`.chomp
+    unless File.exist?(fpc_certs_tar) == true
+       puts "#{FOREMAN_PROXY} certs tar file is not present on the system in path #{fpc_certs_tar} please move the file back to that location or generate a new one on the main server."
+       exit(-1)
+    end
+    backup_configs = base_configs.push(fpc_certs_tar)
+  else
+    backup_configs = base_configs + katello_configs
+  end
+
+  feature_configs = []
+  feature_configs.push("/var/lib/tftpboot") if feature_included?("tftp")
+  feature_configs += ["/var/named/", "/etc/named*"] if feature_included?("dns")
+  feature_configs += ["/var/lib/dhcpd", "/etc/dhcp"] if feature_included?("dhcp")
+  feature_configs.push("/usr/share/xml/scap") if feature_included?("openscap")
+
+  backup_configs + feature_configs
 end
 
 def run_cmd(command, exit_codes=[0])
@@ -100,7 +147,7 @@ def create_directories(directory)
   @dir = File.join directory, "katello-backup-" + DateTime.now.strftime('%Y%m%d%H%M%S')
   puts "Creating backup folder #{@dir}"
   FileUtils.mkdir_p @dir
-  FileUtils.chown_R nil, 'postgres', directory
+  FileUtils.chown_R nil, 'postgres', directory if @databases.include? "pgsql"
   FileUtils.chmod_R 0770, directory
 end
 
@@ -224,15 +271,31 @@ def offline_backup(database, dir_path = nil)
   end
 end
 
+def compress_files
+  psql = spawn('gzip', 'pgsql_data.tar') if @databases.include? "pgsql"
+  mongo = spawn('gzip', 'mongo_data.tar')
+  Process.wait(psql) if @databases.include? "pgsql"
+  Process.wait(mongo)
+end
+
+def plugin_list
+  if @is_foreman_proxy_content
+    JSON.parse(`curl -k https://$(hostname):9090/features`)
+  else
+    plugins = []
+    plugin_list = `foreman-rake plugin:list | grep 'Foreman plugin: '`.lines
+    plugin_list.each do |line|
+      plugin = line.split
+      plugins << "#{plugin[2].chop}-#{plugin[3].chop}"
+    end
+    plugins
+  end
+end
+
 def generate_metadata
   puts "Generating metadata ... "
   os_version = `cat /etc/redhat-release`.chomp
-  plugin_list = `foreman-rake plugin:list | grep 'Foreman plugin: '`.lines
-  plugins = []
-  plugin_list.each do |line|
-    plugin = line.split
-    plugins << "#{plugin[2].chop}-#{plugin[3].chop}"
-  end
+  plugins = plugin_list
   rpms = `rpm -qa`.split("\n")
   system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms}
   File.open('metadata.yml', 'w') do |metadata_file|
@@ -251,23 +314,19 @@ def backup_config_files
   puts "Done."
 end
 
-def compress_files
-  psql = spawn('gzip', File.join(@dir, 'pgsql_data.tar'))
-  mongo = spawn('gzip', File.join(@dir, 'mongo_data.tar'))
-  Process.wait(psql)
-  Process.wait(mongo)
-end
-
 if @dir.nil?
   puts "**** ERROR: Please specify an export directory ****"
   puts optparse
   exit(-1)
-else
-  puts "Starting backup: #{Time.now}"
-  create_directories(@dir.dup)
-  FileUtils.cd @dir
+end
+
+puts "Starting backup: #{Time.now}"
+create_directories(@dir.dup)
+CONFIGS = configure_configs
+
+Dir.chdir(@dir) do
   generate_metadata
-  FileUtils.cp Dir.glob(File.join(@options[:incremental], '.*.snar')), @dir if @options[:incremental]
+  FileUtils.cp(Dir.glob(File.join(@options[:incremental], '.*.snar')), @dir) if @options[:incremental]
   backup_config_files
 
   if @options[:logical_backup]
@@ -284,17 +343,14 @@ else
       online_backup(database)
     end
   else
-    FileUtils.cd("/")
     `katello-service stop #{EXCLUDED}`
     @databases.each do |database|
       offline_backup(database)
     end
     `katello-service start #{EXCLUDED}`
-    FileUtils.cd @dir
     compress_files
   end
-
-  puts "Done with backup: #{Time.now}"
-  puts "**** BACKUP Complete, contents can be found in: #{@dir} ****"
 end
 
+puts "Done with backup: #{Time.now}"
+puts "**** BACKUP Complete, contents can be found in: #{@dir} ****"
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -7,8 +7,88 @@ require 'yaml'
 
 raise 'Must run as root' unless Process.uid == 0
 
+def disable_system_check_option?
+  katello_installer_rpm = `rpm -qa | grep katello-installer`
+  katello_installer_version = katello_installer_rpm[/(\d+\.)(\d+\.)(\d+)/]
+  Gem::Version.new(katello_installer_version) >= Gem::Version.new("3.2.0")
+end
+
+def get_default_program
+  case DEFAULT_SCENARIO
+  when "katello"
+    return "foreman"
+  when FOREMAN_PROXY_CONTENT
+    return "foreman"
+  else
+    return DEFAULT_SCENARIO
+  end
+end
+
+# keep as variables for easy changing in backports
+FOREMAN_PROXY_CONTENT = "foreman-proxy-content"
 DEFAULT_SCENARIO = File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
-DEFAULT_PROGRAM = DEFAULT_SCENARIO == "katello" ? "foreman" : DEFAULT_SCENARIO
+DEFAULT_PROGRAM =  get_default_program
+@proxy = "Foreman Proxy"
+@plural_proxy = "Foreman Proxies"
+
+@options = {}
+@options[:program] = DEFAULT_PROGRAM
+@options[:scenario] = DEFAULT_SCENARIO
+@options[:system_check] = false
+@foreman_proxy_content = @options[:scenario] == FOREMAN_PROXY_CONTENT
+
+opt_parser = OptionParser.new do |opt|
+  opt.banner = "Usage: katello-change-hostname HOSTNAME [OPTIONS]"
+  opt.separator  ""
+  opt.separator  "Example:"
+  opt.separator  " katello-change-hostname foo.example.com -u admin -p changeme"
+  opt.separator  ""
+  opt.separator  "Options"
+
+  opt.on("-u","--username USERNAME","admin username (required)") do |username|
+    @options[:username] = username
+  end
+
+  opt.on("-p","--password PASSWORD","admin password (required)") do |password|
+    @options[:password] = password
+  end
+
+  opt.on("-g","--program PROGRAM","name of the program you are modifying (defaults to #{DEFAULT_PROGRAM})") do |program|
+    @options[:program] = program
+  end
+
+  opt.on("-S","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
+    @options[:scenario] = scenario
+  end
+
+  if disable_system_check_option?
+    opt.on("-d","--disable-system-checks","runs the installer with --disable-system-checks") do |system_check|
+      @options[:system_check] = true
+    end
+  end
+
+  if @foreman_proxy_content
+    opt.on("-c",
+           "--certs-tar CERTS_TAR",
+           "the path to the certs tar generated on the #{DEFAULT_PROGRAM} server with the new hostname (required for #{@plural_proxy})") do |certs_tar|
+      @options[:certs_tar] = certs_tar
+    end
+  end
+
+  opt.on("-h","--help","help") do
+    puts opt_parser
+    exit
+  end
+end
+opt_parser.parse!
+
+def hammer_ping_success
+  return true if @foreman_proxy_content
+  hammer_ping = `hammer ping`
+  success = $?.success?
+  service_failed = hammer_ping.include? "FAIL"
+  success && !service_failed
+end
 
 def is_rhel_version?(version)
   `cat /etc/redhat-release`[/(\d+\.)/].include?(version)
@@ -24,13 +104,7 @@ def get_hostname
   else
     hostname = Socket.gethostname
   end
-  hostname.chomp 
-end
-
-def has_disable_system_check_option?
-  katello_installer_rpm = `rpm -qa | grep katello-installer`
-  katello_installer_version = katello_installer_rpm[/(\d+\.)(\d+\.)(\d+)/]
-  Gem::Version.new(katello_installer_version) >= Gem::Version.new("3.2.0")
+  hostname.chomp
 end
 
 def yesno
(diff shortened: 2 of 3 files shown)


cfouant:
@johnpmitsch - aren't we calling capsule 'foreman-proxy-content' now? I definitely would not have a `--capsule` flag if that is the case

johnpmitsch:
@cfouant yes, that is true, I can change it. I'll add it to the to-do

cfouant:
@johnpmitsch - jumped the gun, just noticed the WIP tag, so thanks!

johnpmitsch:
### To Review ###
_(fpc = foreman-proxy w/ content)_

We are going to essentially "clone" a fpc using the  hostname change and katello-restore
1. Spin up katello + external fpc (will call this fpc1)
2. Associate fpc to organization/lifecycle env and sync some repositories to it
3. Run katello-backup on fpc1
3. Spin up a new external fpc, don't sync (will call this new machine fpc2)
4. Copy over certs tar that was generated for fpc1 (should still be on katello server) to fpc2
1. Run katello-change-hostname to change fpc2 to the hostname of fpc1, you will have to use --certs-tar option and specify fpc1 certs tar
3. Copy over fpc1 backup to fpc2 and run katello-restore
3. Make sure content is actually restored on fpc2 using pulp-admin
4. Delete entry for fpc2 in katello and change /etc/hosts to point fpc1 hostname to fpc2 ip address 
4. Sync fpc2 to make sure all content and functionality is restored
5. Sync new repos to fpc2, delete repos and re-sync, check content is there with pulp-admin
6. Register host to fpc2 and make sure content can be delivered to host

### Testing Scenario Checklist ###
Since quite a bit is refactored in this PR, we have to test the old functionality is still there, here is a list of new and old scenarios to test, please check them off as they are completed and feel free to add more:
- [x] fpc clone using standard backup (using above process) (checked by john - worked fine)
- [x] fpc clone using online backup (using above process) (checked by andrew -worked fine)
- [x] katello-change-hostname works as standalone operation (can change hostname of single fpc with new certs tar) (checked by john)
- [x] katello-restore works as a standalone operation (can restore content on same fpc as backup) (checked by john)
- [ ] standard katello backup and restore still works (checked by john, cfoo is double-checking)
- [ ] online katello backup and restore still works (checked by john, cfoo is double-checking)
- [x] provisioning features work fine (checked by john, can confirm that the correct files are restored based on feature specified)

reminder: snapshots are your friend! :smile:

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+                      '/opt/puppetlabs/puppet/ssl/',
+                      '/var/lib/puppet/foreman_cache_data',
+                      '/var/lib/puppet/ssl',
+                      "/var/lib/tftpboot",
+                      "/etc/dhcp",
+                      "/var/lib/dhcpd",
+                      "/etc/named*",
+                      "/var/named",
+                      "/usr/share/xml/scap"
+
```
ehelms: The indentation here seems strange/off.

katello/katello-backup
```diff
@@ -7,14 +7,17 @@ require "yaml"
...
+@options[:features] = []
 
 EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
 
 optparse = OptionParser.new do |opts|
   opts.banner = "Usage: katello-backup /path/to/dir [options]\n eg: $ katello-backup /tmp/katello-backup"
 
   opts.on("--skip-pulp-content", "Create backup without Pulp content for debugging only") do |config_only|
-    @options[:config_only] = config_only
+    @databases.delete 'pulp'
```
ehelms: This should only skip `/var/lib/pulp` from being backed up not the mongodb database.

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+                      "/etc/candlepin",
+                      "/etc/foreman",
+                      "/etc/hammer",
+                      "/etc/puppetlabs",
+                      "/etc/sysconfig/tomcat*",
+                      "/etc/tomcat*",
+                      "/var/lib/candlepin",
+                    ]
+  if @is_foreman_proxy_content
+    capsule_certs_tar = `cat /etc/foreman-installer/scenarios.d/foreman-proxy-content-answers.yaml | grep certs_tar | awk '{print $2}'`.chomp
```
ehelms: foreman_proxy_certs_tar is what I think you meant here.

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+    unless File.exist?(capsule_certs_tar) == true
+       puts "Capsule certs tar file is not present on the system in path #{capsule_certs_tar} please move the file back to that location or generate a new one on the main server."
```
ehelms: No more Capsules

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+    backup_configs = base_configs.push(capsule_certs_tar)
+  else
+    backup_configs = base_configs + katello_configs
+  end
+  @options[:features].map(&:downcase)
+  feature_configs = []
+  feature_configs.push("/var/lib/tftpboot") if @options[:features].include?("tftp") || @options[:features].include?("all")
+  feature_configs += ["/var/named/", "/etc/named*"] if @options[:features].include?("dns") || @options[:features].include?("all")
+  feature_configs += ["/var/lib/dhcpd", "/etc/dhcp"] if @options[:features].include?("dhcp") || @options[:features].include?("all")
+  feature_configs.push("/usr/share/xml/scap") if @options[:features].include?("openscap") || @options[:features].include?("all")
```
ehelms: You could DRY up these couple lines by making a utility method to check whats included in `options[:features]`

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+  katello_configs = [
+                      "/etc/candlepin",
+                      "/etc/foreman",
+                      "/etc/hammer",
+                      "/etc/puppetlabs",
+                      "/etc/sysconfig/tomcat*",
+                      "/etc/tomcat*",
+                      "/var/lib/candlepin",
+                    ]
+  if @is_foreman_proxy_content
```
ehelms: Consider a blank line above this to separate logical chunks.

katello/katello-backup
```diff
@@ -47,6 +55,64 @@ optparse = OptionParser.new do |opts|
...
+    capsule_certs_tar = `cat /etc/foreman-installer/scenarios.d/foreman-proxy-content-answers.yaml | grep certs_tar | awk '{print $2}'`.chomp
+    unless File.exist?(capsule_certs_tar) == true
+       puts "Capsule certs tar file is not present on the system in path #{capsule_certs_tar} please move the file back to that location or generate a new one on the main server."
+       exit(-1)
+    end
+    backup_configs = base_configs.push(capsule_certs_tar)
+  else
+    backup_configs = base_configs + katello_configs
+  end
+  @options[:features].map(&:downcase)
```
ehelms: Consider a blank line above this to separate logical chunks.

katello/katello-backup
```diff
@@ -59,70 +125,77 @@ def create_directories(directory)
...
   plugin_list.each do |line|
     plugin = line.split
     plugins << "#{plugin[2].chop}-#{plugin[3].chop}"
   end
+  plugins
+end
+
+def generate_metadata
+  os_version = `cat /etc/redhat-release`.chomp
+  plugins = @is_foreman_proxy_content ? "none" : plugin_list
```
ehelms: We should include a list of smart proxy plugins if on the main server and on an external proxy. This will help identify what features are present.

katello/katello-backup
```diff
@@ -7,14 +7,17 @@ require "yaml"
 
 @options = {}
 @dir = nil
+@is_foreman_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
+@databases = ['pulp', 'postgres', 'mongodb']
```
ehelms: I find `pulp` being a database to be quite confusing here given `mongodb` is the Pulp database. Keeping the Pulp content logically separated is a useful readability for understanding the script in my opinion.

katello/katello-change-hostname
```diff
@@ -7,8 +7,88 @@ require 'yaml'
 
 raise 'Must run as root' unless Process.uid == 0
 
+def has_disable_system_check_option?
```
ehelms: The `has_` is redundant with `?` in typical ruby parlance.

katello/katello-change-hostname
```diff
@@ -7,8 +7,88 @@ require 'yaml'
...
+
+  opt.on("-g","--program PROGRAM","name of the program you are modifying (defaults to #{DEFAULT_PROGRAM})") do |program|
+    @options[:program] = program
+  end
+
+  opt.on("-S","--scenario SCENARIO","name of the scenario you are modifying (defaults to #{DEFAULT_SCENARIO})") do |scenario|
+    @options[:scenario] = scenario
+  end
+
+
```
ehelms: Extra line here

katello/katello-change-hostname
```diff
@@ -50,108 +124,132 @@ def yesno
...
+  STDOUT.puts "Checking for certs tarball"
+  if @options[:certs_tar]
+    if File.file?(@options[:certs_tar])
+      true
+    else
+      fail_with_message("#{@options[:certs_tar]} does not exist! Please check the file path and try again")
+    end
+  else
+    fail_with_message("You must specify --certs-tar argument when on a #{@proxy}." \
+                      " These can be generated on the #{DEFAULT_PROGRAM} server using capsule-certs-generate and copied to this machine.")
```
ehelms: foreman-proxy-certs-generate

katello/katello-change-hostname
```diff
@@ -50,108 +124,132 @@ def yesno
...
 
-opt_parser = OptionParser.new do |opt|
-  opt.banner = "Usage: katello-change-hostname HOSTNAME [OPTIONS]"
-  opt.separator  ""
-  opt.separator  "Example:"
-  opt.separator  " katello-change-hostname foo.example.com -u admin -p changeme"
-  opt.separator  ""
-  opt.separator  "Options"
+def get_fpc_answers
+  fpc_answers = YAML.load_file("/etc/foreman-installer/scenarios.d/#{FOREMAN_PROXY_CONTENT}-answers.yaml")
```
ehelms: You could just do the same as you did here -- https://github.com/Katello/katello-packaging/pull/441/files#diff-ba27c55c1473159e75d3f5e6231c8683R29

katello/katello-backup
johnpmitsch: how does one get a  list of plugins while on an external proxy?

katello/katello-backup
johnpmitsch: that is the current behavior of katello-backup, `--skip-pulp-content` will only exclude the very large `/var/lib/pulp` tarball and include everything else.

katello/katello-backup
johnpmitsch: I get the variable `@databases` is a tad misleading, but it's less about categorizing the literal databases and more about categorizing the three areas we want to back up and make them easy to remove in certain scenarios. Maybe we could rename to `@backups`?

katello/katello-change-hostname
johnpmitsch: I'm not sure I understand why? If I'm getting the foreman-proxy answers, shouldn't I specify that file?

katello/katello-change-hostname
```diff
@@ -50,108 +124,132 @@ def yesno
...
+end
 
-  opt.on("-u","--username USERNAME","admin username (required)") do |username|
-    options[:username] = username
-  end
+def succesful_hostname_change_message
+  STDOUT.puts "**** Hostname change complete! ****"
+  if @foreman_proxy_content
+    STDOUT.puts "IMPORTANT: You will have to update the Name and URL of the Smart Proxy in #{@options[:program].capitalize} to the new hostname."
+  else
```
ehelms: This `else` block applies to smart proxies as well since clients can register through them.

katello/katello-change-hostname
```diff
@@ -50,108 +124,132 @@ def yesno
...
   end
+  exit(true)
+end
 
-  opt.on("-h","--help","help") do
-    puts opt_parser
-    exit
+def warning_message
+  STDOUT.print("This script will modify your system.")
+  unless @foreman_proxy_content
```
ehelms: Clients can register through a smart proxy as well and would need to do the update below.

katello/katello-change-hostname
```diff
@@ -50,108 +124,132 @@ def yesno
...
+
+  fpc_installer_args = " --foreman-proxy-content-parent-fqdn #{parent_fqdn}" \
+  " --foreman-proxy-register-in-foreman #{register_in_foreman}" \
+  " --foreman-proxy-foreman-base-url #{base_url}" \
+  " --foreman-proxy-oauth-consumer-key #{oauth_consumer_key}" \
+  " --foreman-proxy-oauth-consumer-secret #{oauth_consumer_secret}" \
+  " --foreman-proxy-content-pulp-oauth-secret #{pulp_oauth_secret}" \
+  " --foreman-proxy-content-certs-tar #{certs_tar}" \
+  " --puppet-server-foreman-url #{foreman_url}" \
+  " --foreman-proxy-register-in-foreman false"
```
ehelms: Why are all these needed in a hotsname change for proxies? The answers file should have the stored previous values for all of these and I would not expect them to change.

katello/katello-restore
```diff
@@ -31,39 +37,69 @@ end
...
+  `runuser - postgres -c "psql -f #{DIR}/pg_globals.dump postgres 2>/dev/null"`
+  `runuser - postgres -c "pg_restore -C -d postgres #{DIR}/foreman.dump"`
+  `runuser - postgres -c "pg_restore -C -d postgres #{DIR}/candlepin.dump"`
+  `service postgresql stop`
+  puts "Done."
+end
+
+def migrate_pulp
+  puts "Migrating pulp databases"
+  `systemctl start mongod qpidd`
```
ehelms: Are we assured all services are stopped and then only these two are getting started? Newer Pulp requires that workers/celery beat and resource manager are running when migrating.

katello/katello-backup
ehelms: Right - the confusing part is when reading the code. I think https://github.com/Katello/katello-packaging/pull/441#discussion_r120391216 explains my concern better.

katello/katello-backup
ehelms: Not sure if there is a rake task to query that or if you will just need to check installed RPMs.

katello/katello-backup
ehelms: There are only three places we backup? I thought we also backed up configuration?

johnpmitsch: @ehelms configs are always included in every backup, so they weren't included in this array

johnpmitsch: @ehelms I guess your concern is having both mongodb and pulp listed here? Maybe we could call `pulp` `pulp-content` or something like that?

katello/katello-restore
johnpmitsch: this will be running after the restore, which happens after katello-service stop in the "restore" method below

katello/katello-backup
johnpmitsch: updated to use the /features endpoint on the proxy

akofink:
Online backup/restore using the steps provided works great. Confirmed on Centos 7 running Katello 3.4 with a RHEL7 client. APJ from me!

johnpmitsch:
@ehelms up for re-review, I tried to keep the newer changes in a separate commit (will squash at the end)

katello/katello-backup
```diff
@@ -35,6 +41,11 @@ optparse = OptionParser.new do |opts|
     @options[:logical_backup] = logical
   end
 
+  opts.on("--features foo,bar", Array, "#{FOREMAN_PROXY.capitalize} features to include in the backup, please specify a list with commas. " \
```
cfouant: Can we leave the `foo,bar` out? and/or only show for foreman proxies?

katello/katello-backup
```diff
@@ -7,14 +7,20 @@ require "yaml"
 
 @options = {}
 @dir = nil
+@is_foreman_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
+@databases = ['pulp', 'postgres', 'mongodb']
+@options[:features] = []
```
cfouant: Options shouldn't be declared outside the option parser. If it is failing elsewhere, perhaps add a `if @options[:features]` tag on failing lines of code?

katello/katello-backup
```diff
@@ -47,6 +58,63 @@ optparse = OptionParser.new do |opts|
...
+    unless File.exist?(fpc_certs_tar) == true
+       puts "#{FOREMAN_PROXY} certs tar file is not present on the system in path #{fpc_certs_tar} please move the file back to that location or generate a new one on the main server."
+       exit(-1)
+    end
+    backup_configs = base_configs.push(fpc_certs_tar)
+  else
+    backup_configs = base_configs + katello_configs
+  end
+
+  @options[:features].map(&:downcase)
```
cfouant: I would put this line inside the OptionParser where it is initially set above (currently line 46). Something like:
`@options[:features] = features.map { |f| f.to_s.downcase }`

katello/katello-restore
```diff
@@ -1,24 +1,30 @@
 #!/usr/bin/env ruby
 
+require "socket"
```
cfouant: nitpick: for uniformity, please use single `'` instead of `"`

katello/katello-restore
```diff
@@ -1,24 +1,30 @@
...
 require 'optparse'
 require 'fileutils'
 require 'date'
 
-STANDARD_BACKUP_FILES = ['config_files.tar.gz', 'pulp_data.tar']
+STANDARD_BACKUP_FILES = ['config_files.tar.gz', 'pulp_data.tar*']
```
cfouant, line 8: Why change the pulp_data tar name?

katello/katello-restore
```diff
@@ -1,24 +1,30 @@
...
+FOREMAN_PROXY_CONTENT_BACKUP_FILES = ['mongo_data.tar.gz']
+FOREMAN_PROXY_CONTENT_ONLINE_BACKUP_FILES = ['mongo_dump']
+FOREMAN_PROXY_CONTENT = "foreman-proxy-content" # keep this variable easy to change for backporting
 
 confirmed = false
+@is_foremnan_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
+@skip_register = false
 
-optparse = OptionParser.new do |opts|
+@optparse = OptionParser.new do |opts|
```
cfouant: Why change this to an instance variable? It is already accessible to the entire script.

katello/katello-restore
```diff
@@ -31,39 +37,72 @@ end
...
+  if @is_foremnan_proxy_content
+    `yes | foreman-installer --scenario #{FOREMAN_PROXY_CONTENT} --reset --foreman-proxy-register-in-foreman false --disable-system-checks`
+  else
+    `yes |foreman-installer --scenario katello --reset`
+  end
   puts "Done.\n"
 end
 
+def restore_psql_dumps
+  puts "Restoring postgresql dump files"
```
cfouant: Nitpick: either change all references printed in messages to `postgresql` or this mention to `postgres` for uniformity.

katello/katello-restore
```diff
@@ -80,10 +119,97 @@ def restore
...
+    false
+  end
+end
+
+def valid_standard_backup
+  if @mongo_data_exists && @pgsql_data_exists
+    !@candlepin_dump_exists && !@foreman_dump_exists && !@mongo_dump_exists
+  else
+    false
+  end
```
cfouant: nitpick: this can be more cleanly done as:

`@mongo_data_exists && @pgsql_data_exists && !(@candlepin_dump_exists || @foreman_dump_exists || @mongo_dump_exists)`

OR

`@mongo_data_exists && @pgsql_data_exists && !@candlepin_dump_exists && !@foreman_dump_exists && !@mongo_dump_exists`

cfouant (changes requested):
@johnpmitsch - great job, the functionality works well! Please see my code review commentary to discuss and/or change. Thanks!

katello/katello-backup
johnpmitsch: I initialized it outside of the option parser so it won't be nil [here](https://git.io/vHMoW), it makes the logic a bit easier in that function instead of handling nil. Is there a specific concern?

katello/katello-backup
johnpmitsch: the features will actually be on katello servers as well as they have default foreman proxies. I'll update to something more like `--features [FEATURE_LIST]` similar to `--incremental`

katello/katello-restore
johnpmitsch: I put `---- *pulp_data.tar is optional ----` [here](https://git.io/vHMKT) as its my understanding that you don't need pulp_data.tar to restore? Please correct me if I'm off on this.

katello/katello-restore
johnpmitsch: I think I had to do that when I moved `optparse` into methods, it couldn't find the variable inside the method

katello/katello-restore
johnpmitsch: let me play around with it again, I'm pretty sure there was a good reason that I didn't go the one-liner route, but will try the above options.

johnpmitsch: @cfouant updated the valid_* methods to use the one-liner syntax, it worked fine in my testing, thanks for the suggestion, I'll run some restores to make sure I'm not missing anything as well.

johnpmitsch:
[test]

johnpmitsch:
In the current state of the PR, the user can't restore a logical backup without removing files from it (a logical backup performs both a standard and online backup, so both sets of files are present). The user is presented with this message when trying to restore a logical or incorrect backup
```
---- The given directory does not contain the required files or has too many files ----
---- All backup directories contain: config_files.tar.gz, pulp_data.tar* ----
---- An online backup directory contains: mongo_dump, candlepin.dump, foreman.dump, pg_globals.dump ----
---- An offline backup directory contains: mongo_data.tar.gz, pgsql_data.tar.gz ----
---- *pulp_data.tar is optional ----
---- Please choose a valid backup directory ----
```
They can move files depending on the type of backup they want to restore. Alternatively, we select one way of restoring a logical backup, or we add flags in to give the user a preference of which backup they want to restore when a logical backup is present. Are there any preferences/opinions on which way is best?

cfouant:
I would prefer if we allow a user to keep their backup files in one place. Since doing a logical-db-backup will create both an offline and online backup, I think we should by default use the offline backup files, and if user wants to override and prefer the online files, we could create a flag in katello-restore

jlsherrill:
Could someone define 'logical backup' ?

johnpmitsch:
@jlsherrill it basically dumps all the things
```
--logical-db-backup          Also dump full database schema during offline backup
```
You will have files from a standard and online backup present when using the flag `--logical-db-backup`

johnpmitsch:
@jlsherrill here are the files in a logical backup:
```
[root@centos7-katello-3-4 ~]# ls /backup/katello-backup-logical/
candlepin.dump  config_files.tar.gz  foreman.dump  metadata.yml  mongo_data.tar.gz  mongo_dump  pg_globals.dump  pgsql_data.tar.gz  pulp_data.tar
```

cfouant:
@jlsherrill - Customer requested to add online schema in case the offline db files are corrupted. http://projects.theforeman.org/issues/18925

katello/katello-restore
cfouant: Nope, that's correct. I thought you were adding a wildcard character, I understand now.

jlsherrill:
How about we default to the offline (telling the user we are using the offline version, and how to remove/move files to use the online), and then if there is some corruption they can switch to the online by deleting (or moving the offline files).

This is both simple from a logic perspective, simple from a user perspective, but still lets the user 'choose' if they really have a need to.

johnpmitsch:
@jlsherrill thanks, we'll go with that unless @cfouant has any objections

katello/katello-backup
cfouant: We shouldn't be referring to an option feature if the flag was not set. If you need to check for specific features, I would do something along the lines of 
`features = @options[:features] || []`

katello/katello-restore
cfouant: I don't see where you moved optparse into methods, it looks like the only change with regards to optparse is that you changed the type of variable it is? Could you double check this? It should very easily be available to every method, as it's declared outside of any other method.

cfouant:
@jlsherrill @johnpmitsch I'd prefer a --use-online-files or something along that line instead. This way the user can 1) keep all files together 2) if there was a corruption instance, it would be easier to support and diagnose with both file types available.

katello/katello-backup
cfouant: :+1:

johnpmitsch:
@cfouant That is a good idea, but I would argue that adding that functionality is a bit outside this PR. Since the current functionality when restoring a logical backup is that it will use the standard backup (correct that if I am wrong), @jlsherrill's method will keep things the way they are, but just add output to clarify that a standard backup is being used.

katello/katello-backup
cfouant: @ehelms - we could call `@databases` `@backups` instead?

cfouant:
@johnpmitsch I'm fine to keep it the way it is already with just a note stating the offline backups are being used for restoration; I would not make any notes about moving files around to perform an online backup.

johnpmitsch:
@cfouant +1

katello/katello-restore
johnpmitsch: @cfouant I updated to remove, it seems to be working without @, not sure what I ran into earlier, I'll test some more to be sure :)

katello/katello-backup
johnpmitsch: good point, updated to include a specified_features method to use that logic

katello/katello-backup
johnpmitsch: updated

katello/katello-backup
```diff
@@ -87,6 +72,66 @@ optparse = OptionParser.new do |opts|
...
+                   "/etc/qpid-dispatch",
+                   "/root/ssl-build",
+                   "/var/www/html/pub",
+                   "/etc/squid",
+                   "/etc/puppetlabs",
+                   '/opt/puppetlabs/puppet/cache/foreman_cache_data',
+                   '/opt/puppetlabs/puppet/ssl/',
+                   '/var/lib/puppet/foreman_cache_data',
+                   '/var/lib/puppet/ssl'
+                 ]
```
ehelms: The indentation here seems weird to me, bit of anitpick but this feels more natural:

```
  base_configs = [
     "/etc/foreman-proxy",
     "/etc/httpd",
     .........
     '/var/lib/puppet/foreman_cache_data',
     '/var/lib/puppet/ssl'
 ]
```

katello/katello-backup
```diff
@@ -156,6 +201,7 @@ def get_base_directory(database)
   result
 end
 
+
```
ehelms: Intentional?

katello/katello-backup
```diff
@@ -224,15 +270,31 @@ def offline_backup(database, dir_path = nil)
...
+def compress_files
+  psql = spawn('gzip', 'pgsql_data.tar') if @databases.include? "pgsql"
+  mongo = spawn('gzip', 'mongo_data.tar')
+  Process.wait(psql) if @databases.include? "pgsql"
+  Process.wait(mongo)
+end
+
+def plugin_list
+  if @is_foreman_proxy_content
+    eval(`curl -k https://$(hostname):9090/features`)
```
ehelms: Why do you have to wrap this in eval?

katello/katello-backup
```diff
@@ -262,39 +324,42 @@ if @dir.nil?
...
+end
 
-  if @options[:logical_backup]
-    @databases.each do |database|
-      online_backup(database)
-    end
-  end
+puts "Starting backup: #{Time.now}"
+create_directories(@dir.dup)
+FileUtils.cd @dir
```
ehelms: This format I find hard to follow what operations need to be run inside of `@dir`. I would find this a bit easier to read and understand if the relevant bits were wrapped in `Dir.chdir(@dir) do end` blocks.

katello/katello-backup
```diff
@@ -262,39 +324,42 @@ if @dir.nil?
...
-      offline_backup(database)
-    end
-    `katello-service start #{EXCLUDED}`
-    FileUtils.cd @dir
-    compress_files
+CONFIGS = configure_configs
+
+puts "Generating metadata ... "
+generate_metadata
+FileUtils.cp Dir.glob(File.join(@options[:incremental], '.*.snar')), @dir if @options[:incremental]
```
ehelms: I think this could benefit readability wise by wrapping the parameters in parens.

katello/katello-backup
```diff
@@ -262,39 +324,42 @@ if @dir.nil?
...
+  `katello-service stop #{EXCLUDED}`
+  @databases.each do |database|
+    offline_backup(database)
+  end
+  `katello-service start #{EXCLUDED}`
+  FileUtils.cd @dir
+  compress_files
 end
 
+puts "Done with backup: #{Time.now}"
```
ehelms: I like the start and end times. Makes me wish it spit out other timestamps after critical parts to the backup to give me some idea of progress and what stage it is at (As well as what parts take the longest). Food for thought here or in a future update to create a little log function that spits out a timestamp and a message.

katello/katello-backup
johnpmitsch: @ehelms to return an array, I figured there was no user input being entered, so using `eval` would be ok

here is the output with and without eval
```
irb(main):001:0> eval(`curl -k https://$(hostname):9090/features`)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    58  100    58    0     0    349      0 --:--:-- --:--:-- --:--:--   349
=> ["logs", "pulpnode", "puppet", "puppetca", "templates", "tftp"]
irb(main):002:0> `curl -k https://$(hostname):9090/features`
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    58  100    58    0     0    369      0 --:--:-- --:--:-- --:--:--   371
=> "[\"logs\",\"pulpnode\",\"puppet\",\"puppetca\",\"templates\",\"tftp\"]"
```

katello/katello-change-hostname
```diff
@@ -50,108 +124,105 @@ def yesno
...
-  STDOUT.puts "Hostname change aborted, no changes have been made to your system"
-  exit(false)
+  fail_with_message("Hostname change aborted, no changes have been made to your system")
 end
 
-unless options[:username] && options[:password]
-  STDOUT.puts "Username and/or Password options are missing!"
-  puts opt_parser
-  exit(false)
+unless @options[:username] && @options[:password]
```
ehelms: There are a lot of "conditional" checks sprinkled throughout this lower section that can cause the script to bail out. They might be nice to consolidate into on area of the script, within a function or within a small module/class for readability and clarity.

katello/katello-change-hostname
```diff
@@ -50,108 +124,105 @@ def yesno
...
-installation_media_paths = `hammer -u #{options[:username]} -p #{options[:password]} medium list | grep "#{old_hostname}" | awk '{ print $5 }'`.split("\n")
-installation_media_ids.zip(installation_media_paths).each do |id, path|
-  `hammer -u #{options[:username]} -p #{options[:password]} medium update --id #{id} --path #{path.sub! old_hostname, new_hostname}`   
+unless @foreman_proxy_content
+  STDOUT.puts "Updating default #{@proxy}"
+  default_capsule_id = `hammer -u #{@options[:username]} -p #{@options[:password]} capsule list | grep #{old_hostname} | awk '{ print $1 }'`[0]
+  # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
+  `hammer -u #{@options[:username]} -p #{@options[:password]} capsule update --id #{default_capsule_id} --url https://#{new_hostname}:9090 --new-name #{new_hostname} 2> /dev/null`
+
+  STDOUT.puts "Updating installation media paths"
```
ehelms: I didn't think we created install media anymore. So why do we need to update them? I realize this is from the previous incarnation of the script just curious.

katello/katello-restore
```diff
@@ -31,39 +42,111 @@ end
...
+  `service postgresql stop`
+  puts "Done."
+end
+
+def migrate_pulp
+  puts "Migrating pulp databases"
+  necessary_services = "mongod qpidd"
+  pulp_services = "pulp_celerybeat pulp_workers pulp_resource_manager"
+  `systemctl start #{necessary_services}`
+  `systemctl stop #{pulp_services}`
```
ehelms: We bake some specialness into katello-service regarding each service for starting and stopping them. I lean towards it being safer to call it using the `only` option.

ehelms: For the above ocmment, apply it to all instances of start/stop throughout (or consider it at least.)

katello/katello-backup
johnpmitsch: +1, I'll leave out of this PR because of its complexity, but will look into adding more timestamps/logging in the near future

katello/katello-change-hostname
johnpmitsch: I also think it's fine to remove, I took it out

katello/katello-backup
ehelms: @johnpmitsch I suppose thats a "safe" assumption, but the array should be valid JSON that you could parse?

johnpmitsch:
@ehelms updated according to your comments

cfouant:
@johnpmitsch - initial tests including snapshots are working! I don't think you'll need to hide the snapshot feature!

johnpmitsch:
@cfouant Snapshots on a foreman-proxy are working on my end too :+1: I'm running tests again to make sure nothing was broken in the small updates I have made, other than that, waiting on you and @ehelms to approve!

cfouant:
I'm getting the error of 'too many files' when trying to restore a backup created with --logical-db-backup:
```[root@capsule ~]# restoretest /tmp/logical/katello-backup-20170620155319/
There are too many files in the backup directory, please check only the required files are present.
---- The given directory does not contain the required files or has too many files
---- All backup directories contain: config_files.tar.gz, pulp_data.tar*
---- A foreman proxy content backup directory contains: mongo_data.tar.gz
---- A foreman proxy content backup directory contains: mongo_dump
---- A logical backup directory contains: mongo_dump, candlepin.dump, foreman.dump, pg_globals.dump, mongo_data.tar.gz, pgsql_data.tar.gz
---- *pulp_data.tar is optional
---- Please choose a valid backup directory
/usr/bin/restoretest:179:in `display_backup_options': undefined local variable or method `optparse' for main:Object (NameError)
	from /usr/bin/restoretest:119:in `backup_invalid_message'
	from /usr/bin/restoretest:213:in `backup_valid?'
	from /usr/bin/restoretest:239:in `<main>'

**Comment by cfouant** [comment]
Also, I see what you were saying about needing to instantiate optparse in katello-restore. I will look into that, because it should be available to that method.

File: katello/katello-restore [hunk 35]
```diff
@@ -1,21 +1,32 @@
...
+FOREMAN_PROXY_CONTENT_BACKUP_FILES = ['mongo_data.tar.gz']
+FOREMAN_PROXY_CONTENT_ONLINE_BACKUP_FILES = ['mongo_dump']
+FOREMAN_PROXY_CONTENT = "foreman-proxy-content" # keep this variable easy to change for backporting
 
 confirmed = false
+@disable_system_checks = false
+@is_foremnan_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
+@skip_register = false
 
 optparse = OptionParser.new do |opts|
```
cfouant: My apologies. This will need to be changed to an instance variable. I didn't realize the scope gate also involved method definitions.

cfouant:
ACK from me, pending the update to the restore method (we agreed above to instead of reject a backup directory with 'too many files' to use the offline files as preference with a note to the user), and making optparse an instance variable again (my bad). I've tested all current methods on a capsule and on a main katello server. Works great! Thanks so much for your work on this!

johnpmitsch:
@cfouant I'm able to run a restore of a logical backup, can you make sure the branch is up to date and if so, do an ls of the directory to see if there is any difference?
```
[root@centos7-katello-3-4 katello-packaging]# ls /backup/katello-backup-20170621134718 
candlepin.dump  config_files.tar.gz  foreman.dump  metadata.yml  mongo_data.tar.gz  mongo_dump  pg_globals.dump  pgsql_data.tar.gz  pulp_data.tar
[root@centos7-katello-3-4 katello-packaging]# ./katello/katello-restore /backup/katello-backup-20170621134718/
WARNING: This script will drop and restore your database.
Your existing installation will be replaced with the backup database.
Once this operation is complete there is no going back.
Are you sure(Y/N)? y
Starting restore from /backup/katello-backup-20170621134718: 2017-06-21 13:52:59 +0000
<snip>
Logical backup detected, using the standard backup files to restore
<snip>
Done with restore: 2017-06-21 13:58:10 +0000
```

johnpmitsch:
@cfouant updated `optparse`

ehelms (approved):

katello/katello-backup
```diff
@@ -251,23 +314,20 @@ def backup_config_files
...
-  create_directories(@dir.dup)
-  FileUtils.cd @dir
+end
+
+puts "Starting backup: #{Time.now}"
+create_directories(@dir.dup)
+CONFIGS = configure_configs
+
+Dir.chdir(@dir) do
+  puts "Generating metadata ... "
```
cfouant: Remove this line (it's inside the generate_metadata method)

katello/katello-restore
```diff
@@ -1,24 +1,35 @@
...
+STANDARD_BACKUP_FILES = ['config_files.tar.gz', 'pulp_data.tar*']
 ONLINE_BACKUP_FILES = ['mongo_dump', 'candlepin.dump', 'foreman.dump', 'pg_globals.dump']
 OFFLINE_BACKUP_FILES = ['mongo_data.tar.gz', 'pgsql_data.tar.gz']
+FOREMAN_PROXY_CONTENT_BACKUP_FILES = ['mongo_data.tar.gz']
+FOREMAN_PROXY_CONTENT_ONLINE_BACKUP_FILES = ['mongo_dump']
+FOREMAN_PROXY_CONTENT = "foreman-proxy-content" # keep this variable easy to change for backporting
 
 confirmed = false
+@disable_system_checks = false
+@is_foremnan_proxy_content = !`rpm -qa | grep foreman-proxy-content`.empty?
```
cfouant: Misspelling, intentional? If not, let's correct.

katello/katello-restore
```diff
@@ -31,39 +42,115 @@ end
...
+end
+
+def migrate_pulp
+  puts "Migrating pulp databases"
+  necessary_services = "mongod,qpidd"
+  pulp_services = "pulp,celerybeat,pulp_workers,pulp_resource_manager"
+  `katello-service start --only #{necessary_services}`
+  `katello-service stop --only #{pulp_services}`
+  `sudo -u apache pulp-manage-db`
+  `katello-service start`
```
cfouant: Let's remove this here. We restart services after the restore_databases method is called, no need to start services 2x.

katello/katello-backup
```diff
@@ -75,6 +56,11 @@ optparse = OptionParser.new do |opts|
     @options[:snapshot_size] = size
   end
 
+  opts.on("--features FEATURES", Array, "#{FOREMAN_PROXY.capitalize} features to include in the backup, please specify a list with commas. " \
```
cfouant, line 59: wrap this option please:
`if @is_foreman_proxy_content`

johnpmitsch: @cfouant These features are present on the main katello server as well

johnpmitsch:
@cfouant updated to remove the unnecessary `katello-service start`, all retesting is passing on my end

katello/katello-restore
```diff
@@ -31,39 +42,114 @@ end
...
   end
+  if !@mongo_data_exists && !@pgsql_data_exists && !valid_logical_backup
+    if @foreman_dump_exists && @candlepin_dump_exists
+      restore_psql_dumps
+    end
+    if @mongo_dump_exists
+      restore_mongo_dump
+    end
+  end
+  migrate_pulp
```
cfouant, line 146: Is this necessary in every case?

johnpmitsch: I've found that sometimes pulp needs to be migrated after a restore, especially if restoring from a different katello z-stream version. It won't change anything if pulp is already migrated,  but is there for extra insurance to make sure the pulp database is up to date after a restore

cfouant:
ACK @johnpmitsch great job, wonderful work!

katello/katello-restore
cfouant: Ok, that's wise to keep in a restore then!

PR #443, opened 2017-05-09
Fixes #19499 - Convert metadata file to yml format

johnpmitsch:
This also adds a package list to the metadata file,
which will be helpful to users or scripts in
determining versions

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -3,6 +3,7 @@
 require 'optparse'
 require 'fileutils'
 require 'date'
+require "yaml"
 
 @options = {}
 @dir = nil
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -106,6 +107,20 @@ def compress_files
...
+  os_version = `cat /etc/redhat-release`.chomp
+  plugin_list = `foreman-rake plugin:list | grep 'Foreman plugin: '`.lines
+  plugins = []
+  plugin_list.each do |line|
+    plugin = line.split
+    plugins << "#{plugin[2].chop}-#{plugin[3].chop}"
+  end
+  rpms = `rpm -qa`.split("\n")
+  system_facts = {os_version: os_version, plugin_list: plugins, rpms: rpms}
+  metadata.puts system_facts.to_yaml
```
jlsherrill: make sure to close the file:
`metadata.close`.

You might also split out the plugin list fetching into its own method.

katello/katello-backup
```diff
@@ -106,6 +107,20 @@ def compress_files
   Process.wait(mongo)
 end
 
+def generate_metadata
+  metadata = File.new 'metadata.yaml', 'w'
```
jlsherrill: nitpick, but i'd use () here.

katello/katello-backup
ehelms: Since you are writing only at the end, you can just do a File.open block here

johnpmitsch:
@ehelms @jlsherrill updated

jlsherrill:
ACK thanks @johnpmitsch

PR #449, opened 2017-05-18
Fixes #19256 - allow older goferd on el5

jlsherrill:


jlsherrill:
@beav, i think this is related to a change you made a while back

beav:
ACK, looks good.

PR #456, opened 2017-05-25
Refs #18874: jquery-ui-rails is no longer required

ehelms:
Based on the removal from the codebase -- http://projects.theforeman.org/issues/18874


jlsherrill:
ACK

PR #457, opened 2017-05-25
fixes #19673 - removes /var/www/pub from pulp_data.tar

cfouant:


ehelms:
ACK - thanks @cfouant !

PR #459, opened 2017-06-01
fixes #18328, #20268 - adds interaction to confirm service shutdown, provides warning for online backup

cfouant:
This adds interaction to confirm before stopping services, or proceed with online backup after reading warning. It also adds a "-y/--assumeyes" flag to bypass interaction by assuming yes to questions.

### DIFF
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -6,6 +6,7 @@ require 'date'
 require 'yaml'
 require 'find'
 require 'json'
+require 'highline/import'
 
 EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
 DATABASES = ['pulp', 'pgsql', 'mongodb']
@@ -63,6 +64,10 @@ optparse = OptionParser.new do |opts|
 
   opts.on("--preserve-directory", "Do not create a time-stamped subdirectory") do |no_subdir|
     @options[:no_subdir] = no_subdir
+  end
+
+  opts.on("-y", "--assumeyes", "Bypass interaction by answering yes") do |confirm|
+    @options[:confirm] = confirm
   end
 
   opts.parse!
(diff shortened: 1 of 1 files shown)


katello/katello-backup
```diff
@@ -63,6 +67,18 @@ def create_directories(directory)
...
+  puts "WARNING: This script will stop your services."
+  print "Are you sure(Y/N)? "
+  response = gets.chomp
+  if /[Y]/i.match(response)
+  else
+    puts "**** cancelled ****"
+    FileUtils.rm_rf @dir
+    exit(-1)
+  end
+end
```
johnpmitsch, line 207: Highline has a simple [agree method](http://www.rubydoc.info/gems/highline/1.6.21/HighLine:agree) that could simplify this check

you can run do something like this
```
require 'highline' (at top)
response = HighLine.agree('WARNING: This script will stop your services. Do you want to proceed(y/n)?')
```

johnpmitsch (changes requested):

johnpmitsch:
Should we make it default to confirm the stopping of services? There could be a `-y/--yes` flag to override. I'm not sure which tactic would be better for users.

cfouant:
@johnpmitsch that would likely break a lot of cron jobs, so we are going this route instead

johnpmitsch:
@cfouant  I am thinking we should go the route of shutting down the services by default and then the user can override with `-y` for scripting. 
From the issue:
> This can have serious consequences if the user does not know ahead of time that the server will be shut down during the backup.

if the user doesn't know that services are going to be shut down, will they really pass a` --confirm-shutdown` flag? To me, it seems like to address the original intent of this issue and user frustration we should default to a confirmation and have a flag to override (similar to a command like `yum install`). 

There is also the benefit of the user taking responsibility for their actions - if they pass with a flag to override the confirmation and there are negative consequences, that is because of their specific intent to bypass a check, not because they didn't know of the action happening. (aka not our fault! :) )

katello/katello-backup
cfouant: Nice! Updated. Also... probably should have something for the true case scenario, guess that would help haha

cfouant:
@johnpmitsch I agree, it seems counterintuitive to add this flag, and doesn't really make a ton of sense. When I first was planning this @ehelms pointed out doing so would break existing cron scripts. @ehelms - your thoughts? @johnpmitsch makes a good point

cfouant:
@johnpmitsch please re-review, PR has been updated to by default get user input before shutting down services.

johnpmitsch (approved):

johnpmitsch:
@cfouant approved, you need a rebase and then it's good to go :)

cfouant:
@johnpmitsch updated!

PR #460, opened 2017-06-01
fixes #19743, #18358 - allows user to bypass subdir creation

cfouant:


johnpmitsch:
Not related to this PR, but it looks like incremental is broken

[root@steak ~]# /home/vagrant/katello-packaging/katello/katello-backup --incremental /backup/no_subdir_with_inc/
katello-backup: **** ERROR: Please specify an export directory ****

Assuming I have the right syntax here, it looks like [this check](https://git.io/vHVT0) is still going to run even though `--incremental` is [using the directory argument](https://git.io/vHVTz)

If this is correct, I can file an issue and fix

johnpmitsch:
The reason I was testing `--incremental` (above comment) is because I am wondering what happens if I run `--incremental` and `--preserve-directory`, will there be a subdirectory created?

johnpmitsch (changes requested):
I'm getting this message when running with --preserve-directory on an existing backup

gzip: pgsql_data.tar.gz already exists; do you wish to overwrite (y or n)? gzip: mongo_data.tar.gz already exists; do you wish to overwrite (y or n)?

This will be an issue for those who are going to use a cron job to backup to the same directory. I am thinking we should remove all files in the backup with a yes/no confirmation that their backup will be overwritten. (that can be overridden with -y/--yes kind of thing).

cfouant:
@johnpmitsch you'd still need to define the backup directory so it would look something like:

`# katello-backup /tmp/backup --incremental /tmp/backup --preserve-directory`

Running --preserve-directory will not create a subdirectory no matter what other options are listed. Using this with --incremental would probably be the most likely scenario.

johnpmitsch:
@cfouant so I could do something like 
`katello-backup /tmp/backup --incremental /tmp/mylargebackup --preserve-directory /tmp/mylargebackup` and do an incremental backup that replaces the base backup?

johnpmitsch:
> I am thinking we should remove all files in the backup with a yes/no confirmation that their backup will be overwritten. (that can be overridden with -y/--yes kind of thing).

Actually reconsidering my suggestion... If we remove the directory and the backup fails, a user could be left backup-less. Maybe we could move the files in the directory to a temporary directory that gets deleted at the end of the script? @ehelms or @jlsherrill may have other ideas (see [here](https://git.io/vH6Ne) for context)

cfouant:
@johnpmitsch I added a '-f' so you should no longer get that error. It will overwrite the existing files without interrupting the backup now.

cfouant:
@johnpmitsch the command would look different. The --preserve-directory flag does not require any arguments.

`# katello-backup <backup directory location> --incremental <existing backup directory to increment from> --preserve-directory`

So in the case of wanting the incremental directory to be the same as the location which you're overwriting, it would look like this:

`# katello-backup /tmp/my_backup_directory --incremental /tmp/my_backup_directory --preserve-directory`

johnpmitsch:
@cfouant I am seeing this when running `/katello-backup --preserve-directory /backup/katello-backup-20170608184218/`
```
gzip: pgsql_data.tar.gz already exists; do you wish to overwrite (y or n)? gzip: mongo_data.tar.gz already exists; do you wish to overwrite (y or n)? y
gzip: f: No such file or directory
```

cfouant:
@johnpmitsch - please rereview, thanks!

cfouant:
[test]

johnpmitsch (approved):

johnpmitsch:
nice change @cfouant!

PR #471, opened 2017-06-19
Remove python-urllib3 in favor of whats in base

ehelms:
The version in base CentOS has caught up so we don't need to carry this anymore.


evgeni:
I don't see https://github.com/Katello/katello-packaging/blob/master/python-urllib3/key-connection-pools-off-custom-keys.patch being part of https://git.centos.org/blob/rpms!python-urllib3/c7/SPECS!python-urllib3.spec?

ehelms:
I don't have a clue what that spec file you linked is from, but EPEL has python-urllib3 1.19.1 which was released in November 2016 (https://dl.fedoraproject.org/pub/epel/7/x86_64/p/python34-urllib3-1.19.1-3.el7.noarch.rpm)

evgeni:
That spec is fron CentOS (not EPEL) git.

ehelms:
Ok, but what's in EPEL is what we care about since that's what a user will
install?

evgeni:
But EPEL has only python34-urllib3, not python-urllib3 (EPEL can't ship newer versions of packages in CentOS/RHEL proper).

PR #483, opened 2017-07-19
fixes #20353 - Run apipie cache after install

parthaa:


ehelms:
The rpm build already runs this step, so I don't get how this would solve the problem. What is the missing piece that makes this work when run at the end ? My first thought is this working around the root issue.

parthaa:
Apipie needs to re run after ostree plugin is installed. We need to update the json for repositories controller.

ehelms:
I think then the ostree RPM should be running this during its build phase. Thats where every other plugin/RPM runs this especially since it can be time consuming its better to do it during RPM build than on a users system.

parthaa:
Added this puppet katello https://github.com/Katello/puppet-katello/issues/201 . Closing the packaging aspect here

PR #487, opened 2017-07-27
Fixes #20382 - Require candlepin 2.0

jlsherrill:


katello/katello.spec
```diff
@@ -58,8 +58,8 @@ Requires: createrepo >= 0.9.9-18%{?dist}
 Requires: squid
 Requires: mod_xsendfile
 
-Requires(post): candlepin
-Requires: candlepin-selinux
+Requires(post): candlepin >= 2.0
+Requires: candlepin-selinux >= 2.0
 Requires: java-openjdk >= 1:1.7.0
 Requires: java-openjdk < 1:1.8.0.45
 Requires: lsof
```
sean797: Out of scope for this PR but `lsof` seems like a weird dependency to me.

sean797 (approved):
Thanks @jlsherrill 👍

jlsherrill:
@sean797 went ahead and removed lsof, i think it was some really old dependency for some script that used to use it.  Does not seem used anymore

ehelms (approved):

PR #488, opened 2017-07-27
Fixes #19281 - Move katello commands to sbin

johnpmitsch:


evgeni:
you need to adapt the install part too:
https://github.com/johnpmitsch/katello-packaging/blob/1625bec3f07aec345ce8de9a194224002d8c4672/katello/katello.spec#L92-L98

johnpmitsch:
@evgeni updated, I don't know if we want katello-service moved as well

evgeni:
Given `katello-service status` works as a user, I'd vote for leaving it in bin. @ehelms?

ehelms:
What user is going to run that? I say keep them all together.

johnpmitsch:
updated

ehelms (approved):
@evgeni what say ye?

evgeni:
I'd say my CDO kicks in online 92 and the fact that the install and files sections are sorted differently, but otherwise I'm good!

PR #489, opened 2017-07-27
Upgrade katello-selinux to 3.0.3

akofink:


ehelms:
@akofink you added the source as abinary and not git annex file.

akofink:
Updated

jlsherrill:
ACK thanks @akofink !

PR #491, opened 2017-07-28
[WIP] Fixes #20430 - katello-change-hostname performs installer dry-run

johnpmitsch:
The installer will do a dry run in the precheck to check for
potential problems before making changes.

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -125,6 +125,22 @@ def fail_with_message(message, opt_parser=nil)
   exit(false)
 end
 
+def move_file(original, backup, reverse=false)
+  STDOUT.puts("original " + original)
+  STDOUT.puts("backup " + backup)
+  base_original = File.dirname(original)
+  base_backup = File.dirname(backup)
+  original_filename = File.basename(original)
+  backup_filename = File.basename(backup)
+  if reverse
+    `mv #{base_backup}/#{backup_filename} #{base_original}`
+  else
+    STDOUT.puts("mv #{base_original}/#{backup_filename} #{base_backup}")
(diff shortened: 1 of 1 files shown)


katello/katello-change-hostname
```diff
@@ -197,6 +208,15 @@ def precheck(response)
     fail_with_message("Please specify a hostname.", opt_parser)
   end
 
+  installer_precheck = installer_command + " -d"
```
ehelms: I think you mean `-n` here?

johnpmitsch:
@ehelms updated

johnpmitsch:
@ehelms which do we want to do?

```
dry run of the installer (exit if it fails)
update hostname on system + remove certs
real run of the installer
```
or
```
update hostname on the system + remove certs
dry run of the installer (exit if it fails)
real run of the installer
```

My concern with the second option is there would be permanent changes. However, with the first option, the dry run of the installer isn't as thorough of a check.

johnpmitsch:
@ehelms 
Apologies for the stream of consciousness, but here is what I'm thinking for the hostname-change pre-check
```
update hostname on system
move certs to files/folders in a directory we create in /tmp
Perform dry run of the installer
if dry run is successful:
  continue with change hostname script, which will run the actual installer
else:
  move certs back to appropriate location
  change system hostname back to the old hostname
  fail out of script w/ error
```
I think this will be the most robust way to check if the installer will succeed when running katello-change hostname while preventing irreversable changes from happening to user's systems. Curious to hear your thoughts

ehelms:
@johnpmitsch planning to revisit this?

johnpmitsch:
I am closing this as the dry run of the installer would fail sometimes and the actual run would pass. I am working on another "cleanup" solution.

PR #499, opened 2017-08-10
Fixes #20550 - support for remote DB in backup and restore

mbacovsky:
Backup and restore scripts were changed to access postgress DB
using the configuration from Foreman and Candlepin and not directly
as DB superuser.

In online backup we skip dump of global objects as usually there are no
privileges to restore them on the remote DB.

In offline backup we do ofline backup as before only the remote DBs
are backed up using pg_dump as in online backup because we do not have
direct access to /var/lib/pgsql/data. We assume the remote DB is always
up.

The restore was updated accordingly and can restore offline backup
including dumps of the remote dbs.

katello-service was fixed to not touch postgresql service when the DB is
remote.

### DIFF
--- a/katello/db_config.rb
+++ b/katello/db_config.rb
@@ -45,5 +45,20 @@ module KatelloUtilities
     def pg_dump_command(config, dump_file)
       "PGPASSWORD='#{config[:password]}' pg_dump -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -Fc #{config[:database]} > #{dump_file}"
     end
+
+    def pg_sql_statement(config, statement)
+      pg_command(config, 'psql', '-t -c "' + statement + '"')
+    end
+
+    # WARNING: deletes all the data from a dabase. No warnings. No confirmations.
+    def empty_database!(config)
+      generate_delete_statements = pg_sql_statement(config, %q(
+        select string_agg('drop table if exists \"' || tablename || '\" cascade;', '')
+        from pg_tables
+        where schemaname = 'public';
+      ))
+      delete_statements = `#{generate_delete_statements}`
+      system(pg_sql_statement(config, delete_statements)) if delete_statements
+    end
   end
 end
--- a/katello/db_config.rb
+++ b/katello/db_config.rb
@@ -38,12 +38,16 @@ module KatelloUtilities
       remote_db?(foreman) || remote_db?(candlepin)
     end
 
+    def pg_command_base(config, command, args)
+      "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} #{args}"
+    end
+
     def pg_command(config, command, args)
-      "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -d #{config[:database]} #{args}"
+      pg_command_base(config, command, "-d #{config[:database]} #{args}")
     end
 
     def pg_dump_command(config, dump_file)
-      "PGPASSWORD='#{config[:password]}' pg_dump -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -Fc #{config[:database]} > #{dump_file}"
+      pg_command_base(config, 'pg_dump', "-Fc #{config[:database]} > #{dump_file}")
     end
 
     def pg_sql_statement(config, statement)
--- a/katello/db_config.rb
+++ b/katello/db_config.rb
@@ -21,8 +21,6 @@ module KatelloUtilities
           :database => answers['katello']['candlepin_db_name'] || 'candlepin',
           :host => answers['katello']['candlepin_db_host'] || 'localhost',
           :port => answers['katello']['candlepin_db_port'] || '5432',
-          :ssl => answers['katello']['candlepin_ssl'] || false,
-          :ssl_verify => answers['katello']['candlepin_ssl_verify'] || false
       }
     end
 
--- a/katello/db_config.rb
+++ b/katello/db_config.rb
@@ -1,0 +1,49 @@
+require 'yaml'
+
+module KatelloUtilities
+  class DBConfig
+    attr_reader :foreman, :candlepin
+    INSTALLER_CONFIG = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
+
+    def initialize(scenario = INSTALLER_CONFIG)
+      config = YAML.load(File.read(scenario))
+      answers = YAML.load(File.read(config[:answer_file]))
+      @foreman = {
+          :username => answers['foreman']['db_username'],
+          :password => answers['foreman']['db_password'],
+          :database => answers['foreman']['db_database'] || 'foreman',
+          :host => answers['foreman']['db_host'] || 'localhost',
+          :port => answers['foreman']['db_port'] || '5432'
+      }
+      @candlepin = {
+          :username => answers['katello']['candlepin_db_user'],
+          :password => answers['katello']['candlepin_db_password'],
+          :database => answers['katello']['candlepin_db_name'] || 'candlepin',
+          :host => answers['katello']['candlepin_db_host'] || 'localhost',
+          :port => answers['katello']['candlepin_db_port'] || '5432',
+          :ssl => answers['katello']['candlepin_ssl'] || false,
+          :ssl_verify => answers['katello']['candlepin_ssl_verify'] || false
+      }
+    end
+
+    def remote_db?(config)
+      !['localhost', '127.0.0.1', `hostname`.strip].include? config[:host]
+    end
+
+    def any_local_db?
+      !remote_db?(foreman) || !remote_db?(candlepin)
+    end
+
+    def any_remote_db?
+      remote_db?(foreman) || remote_db?(candlepin)
+    end
+
+    def pg_command(config, command, args)
+      "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -d #{config[:database]} #{args}"
+    end
+
+    def pg_dump_command(config, dump_file)
+      "PGPASSWORD='#{config[:password]}' pg_dump -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -Fc #{config[:database]} > #{dump_file}"
+    end
+  end
+end
--- a/katello/katello-service
+++ b/katello/katello-service
@@ -2,6 +2,14 @@
 
 require 'optparse'
 
+# check if production install
+if __FILE__.start_with?('/sbin/')
+  UTILS_PATH='/usr/share/katello'
+else
+  UTILS_PATH=File.expand_path('..', __FILE__)
+end
+require File.join(UTILS_PATH, 'db_config.rb')
+
 ACTIONS = ['restart', 'stop', 'start', 'status', 'list', 'enable', 'disable']
 COMMAND = '/usr/sbin/service-wait'
 SERVICES = {
@@ -25,6 +33,7 @@ SERVICES = {
 }
 
 @options = {:excluded => []}
+@db_config = KatelloUtilities::DBConfig.new
 
 OptionParser.new do |opts|
   opts.banner = "Usage: katello-service [options] [#{ACTIONS.join('|')}]"
@@ -50,6 +59,7 @@ OptionParser.new do |opts|
 end
 
 def service_exists?(service)
+  return false if !@db_config.any_local_db? and service == 'postgresql'
   systemd = `systemctl is-enabled #{service} 2>&1`.strip
   systemd == 'enabled' || systemd == 'disabled'
 end
--- a/katello/restore.rb
+++ b/katello/restore.rb
@@ -12,6 +12,7 @@ module KatelloUtilities
       @standard_backup_files = ['config_files.tar.gz', 'pulp_data.tar*']
       @online_backup_files = ['mongo_dump', 'candlepin.dump', 'foreman.dump', 'pg_globals.dump']
       @offline_backup_files = ['mongo_data.tar.gz', 'pgsql_data.tar.gz']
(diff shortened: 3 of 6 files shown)


mbacovsky:
The installer --reset is dropping the DBs and the owner usually do not have the permission to re-create them. I'll add small fix skipping the drop db for remote dbs.

Other then that I was able to backup and restore Katrello completely on remote DB.

It would be great if someone can check that the changes I had to do to the backup procedure make sense and are okay with what is expected by admins in the field.

katello/katello-backup
```diff
@@ -10,11 +10,14 @@ require 'highline/import'
...
 EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
 DATABASES = ['pulp', 'pgsql', 'mongodb']
+FOREMAN_DB_CONFIG = '/etc/foreman/database.yml'
+CANDLEPIN_DB_CONFIG = '/etc/candlepin/candlepin.conf'
 
 @options = {}
 @dir = nil
 @databases = DATABASES.dup
 
+
```
sean797: Unnecessary?

katello/katello-backup
```diff
@@ -10,11 +10,14 @@ require 'highline/import'
 
 EXCLUDED="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
 DATABASES = ['pulp', 'pgsql', 'mongodb']
+FOREMAN_DB_CONFIG = '/etc/foreman/database.yml'
+CANDLEPIN_DB_CONFIG = '/etc/candlepin/candlepin.conf'
```
sean797: Instead of reading these 2 files, I think it would be a lot simpler to read `/etc/foreman-installer/scenarios.d/#{scenario}-answers.yaml` ?
You could get the scenario from `/etc/foreman-installer/scenarios.d/last_scenario.yaml`

katello/katello-restore
```diff
@@ -4,13 +4,17 @@ require 'socket'
...
 
 STANDARD_BACKUP_FILES = ['config_files.tar.gz', 'pulp_data.tar*']
 ONLINE_BACKUP_FILES = ['mongo_dump', 'candlepin.dump', 'foreman.dump', 'pg_globals.dump']
 OFFLINE_BACKUP_FILES = ['mongo_data.tar.gz', 'pgsql_data.tar.gz']
+OFFLINE_REMOTE_BACKUP_FILES = ['mongo_data.tar.gz', 'candlepin.dump', 'foreman.dump']
 FOREMAN_PROXY_CONTENT_BACKUP_FILES = ['mongo_data.tar.gz']
 FOREMAN_PROXY_CONTENT_ONLINE_BACKUP_FILES = ['mongo_dump']
 FOREMAN_PROXY_CONTENT = "foreman-proxy-content" # keep this variable easy to change for backporting
+FOREMAN_DB_CONFIG = '/etc/foreman/database.yml'
+CANDLEPIN_DB_CONFIG = '/etc/candlepin/candlepin.conf'
```
sean797: ditto

katello/katello-service
```diff
@@ -49,7 +50,14 @@ OptionParser.new do |opts|
   @options[:action] = ARGV[0]
 end
 
+def remote_db?
+  config_file = '/etc/foreman/database.yml'
```
sean797: ditto.

katello/katello-restore
```diff
@@ -75,14 +79,47 @@ def reset_katello
...
+      'username' => full_config['org.quartz.dataSource.myDS.user'],
+      'password' => full_config['org.quartz.dataSource.myDS.password'],
+      'database' => uri[4],
+      'host' => uri[2],
+      'port' => uri[3] || '5432'
+  }
+end
+
+def is_remote_db?(host)
+  host != 'localhost'
```
sean797: `127.0.0.1` & `hostname`might also be valid

katello/katello-restore
```diff
@@ -75,14 +79,47 @@ def reset_katello
...
-  run_cmd("runuser - postgres -c 'dropdb foreman'")
-  run_cmd("runuser - postgres -c 'dropdb candlepin'")
-  run_cmd("runuser - postgres -c 'psql -f #{File.join(@dir, "pg_globals.dump")} postgres 2>/dev/null'")
-  run_cmd("runuser - postgres -c 'pg_restore -C -d postgres #{File.join(@dir, "foreman.dump")}'")
-  run_cmd("runuser - postgres -c 'pg_restore -C -d postgres #{File.join(@dir, "candlepin.dump")}'")
+  foreman_config = foreman_db_config
+  if is_remote_db?(foreman_config[:host])
+    candlepin_config = candlepin_db_config
+    run_pg_cmd(foreman_config, 'pg_restore', "--no-privileges --clean -n public #{File.join(@dir, "foreman.dump")}")
+    run_pg_cmd(candlepin_config, 'pg_restore', "--no-privileges --clean -n public #{File.join(@dir, "candlepin.dump")}")
```
sean797: Is it possible it might be in a namespace other than `public` ?

katello/katello-backup
```diff
@@ -299,19 +333,27 @@ def offline_backup(database, dir_path = nil)
...
     end
   when 'pgsql'
+    foreman_config = foreman_db_config
     dir_path ||= '/var/lib/pgsql/data'
-    FileUtils.cd dir_path do
-      puts "Backing up postgres db..."
-      run_cmd("tar --selinux --create --file=#{File.join(@dir, 'pgsql_data.tar')} --listed-incremental=#{File.join(@dir, '.postgres.snar')} --transform 's,^,var/lib/pgsql/data/,S' -S *")
-      puts "Done."
+    if is_remote_db?(foreman_config['host'])
+      puts "Backup of #{dir_path} is not supported for remote databases. Doing online backup instead... "
```
sean797: s/Doing online backup instead/Doing postgres online backup instead/g ? Just to be clear

sean797 (changes requested):
Thanks for doing this @mbacovsky 👍 

Since this doesn't include support for remote mongodb would you mind changing the commit message, just to be clear.
  
Couple of inline comments, generally I don't like the approach of reading '/etc/foreman/database.yml' & '/etc/candlepin/candlepin.conf'. I think the answer file would be a lot simpler to implement. I guess that does give us a dependence on the answer file, I think its fair to say everyone uses the installer to build Katello?

katello/katello-backup
mbacovsky: @sean797 that's a good idea, I'll reimplement this.

katello/katello-restore
mbacovsky: @sean797 probably not, I'll remove the schema restriction

mbacovsky:
Unfortunately I'm not going to finish my updates before I leave for vacation. I'll update 1st week of September.

katello/katello-restore
mbacovsky: @sean797 I had to keep it restricted to the public schema. Otherwise it tried to clean also the procedural language and objects that the user may not be allowed to drop.

mbacovsky:
@sean797, @beav , I've updated this PR. I've extracted the common code and put it to the `/usr/share/foreman/script`. Feel free to suggest better location. I've also addressed the comments above and rebased the PR.
The `katello-restore` requires change in how the --reset feature of the installer approaches the db. The PR will follow soon.

sean797:
@mbacovsky I would prefer if it could go in lib/ like https://github.com/Katello/katello-packaging/pull/524/

Overall looks good 👍  but I haven't tested it.

mbacovsky:
@sean797 I'll wait till the other PR is merged and update the code accordingly. `lib/` is more convenient place for the db handling code.

mbacovsky:
@sean797 I've updated the patch to work with the refactored scripts. I've updated the SPEC and tested the scratch builds with rest of the packages from Katello nightly. I was able to backup and restore my Katello successfully. There were some hiccups during db:migrate and internal proxy check. It was fixed after re-running the installer. I'm not sure if that is a nightly thing or my semi-devel box upgraded from Katello 3.4.

There is one fix in progress to prevent dropping remote DB in reset hook. I'll send it later today.

katello/db_config.rb
```diff
@@ -0,0 +1,49 @@
...
+    def any_remote_db?
+      remote_db?(foreman) || remote_db?(candlepin)
+    end
+
+    def pg_command(config, command, args)
+      "PGPASSWORD='#{config[:password]}' #{command} -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -d #{config[:database]} #{args}"
+    end
+
+    def pg_dump_command(config, dump_file)
+      "PGPASSWORD='#{config[:password]}' pg_dump -U #{config[:username]} -h #{config[:host]} -p #{config[:port]} -Fc #{config[:database]} > #{dump_file}"
```
sean797: DRY. Use `pg_command` methods above ?

katello/db_config.rb
```diff
@@ -0,0 +1,49 @@
...
+          :port => answers['foreman']['db_port'] || '5432'
+      }
+      @candlepin = {
+          :username => answers['katello']['candlepin_db_user'],
+          :password => answers['katello']['candlepin_db_password'],
+          :database => answers['katello']['candlepin_db_name'] || 'candlepin',
+          :host => answers['katello']['candlepin_db_host'] || 'localhost',
+          :port => answers['katello']['candlepin_db_port'] || '5432',
+          :ssl => answers['katello']['candlepin_ssl'] || false,
+          :ssl_verify => answers['katello']['candlepin_ssl_verify'] || false
```
sean797: These 2 SSL options don't seem to actually be used?

sean797 (changes requested):

katello/db_config.rb
mbacovsky: Currently not, but as it is part of the DB configuration I chose to keep it. It may become useful when building URI for JDBC like in https://github.com/Katello/katello-installer/blob/master/hooks/pre/30-upgrade.rb#L39. If it is a problem, I can remove it.

mbacovsky:
@sean797, refactored the pg_command a bit...

katello/restore.rb
```diff
@@ -82,11 +83,19 @@ def reset_katello
...
+          run_cmd("runuser - postgres -c 'psql -f #{File.join(@dir, "pg_globals.dump")} postgres 2>/dev/null'")
+        end
+        run_cmd(db_config.pg_command(db_config.foreman, 'pg_restore', "--no-privileges --clean -n public #{File.join(@dir, "foreman.dump")}"))
+        run_cmd(db_config.pg_command(db_config.candlepin, 'pg_restore', "--no-privileges --clean -n public #{File.join(@dir, "candlepin.dump")}"))
+      else
+        run_cmd("runuser - postgres -c 'dropdb foreman'")
+        run_cmd("runuser - postgres -c 'dropdb candlepin'")
+        run_cmd("runuser - postgres -c 'psql -f #{File.join(@dir, "pg_globals.dump")} postgres 2>/dev/null'")
+        run_cmd("runuser - postgres -c 'pg_restore -C -d postgres #{File.join(@dir, "foreman.dump")}'")
+        run_cmd("runuser - postgres -c 'pg_restore -C -d postgres #{File.join(@dir, "candlepin.dump")}'")
```
sean797, line 97: nitpick. Can these work in a similar way to `backup.rb`? Making use of the methods in `DBConfig`, just for consistency. Your'll probably want to add a `pg_restore_command` method.

https://github.com/Katello/katello-packaging/pull/499/files#diff-e5c9a75a6cce52c183d1d26a2500311eR219

katello/db_config.rb
sean797: Please do, we can add them when we need them. Seem unnecessary to me :)

sean797 (changes requested):
Okay, this is looking really good now! Much cleaner than before and working nicely! Thanks @mbacovsky 👍

2 small comments and then we should be good to merge 🎆

katello/restore.rb
mbacovsky: @sean797, I didn't want to touch the restore of the local DB as it is not functionally equivalent to the commands run as pg admin. However the difference is small and should affect non-standard foreman DB (custom schema changes, privileges etc.).  If you think it is safe we can remove the whole else part and use the same procedure that is used for remote DB. What do you think?

sean797: Makes sense 👍

sean797 (approved):
ACK, thanks @mbacovsky 🎆 

Whoever does the merge, please squash the commits

mbacovsky:
@sean797 thanks for all the reviewing effort!

ehelms:
@cfouant can you give this a once over?

katello/db_config.rb
```diff
@@ -0,0 +1,66 @@
+require 'yaml'
+
+module KatelloUtilities
+  class DBConfig
+    attr_reader :foreman, :candlepin
+    INSTALLER_CONFIG = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
+
+    def initialize(scenario = INSTALLER_CONFIG)
```
ekohl, line 8: This is very tight coupling with the puppet modules and that is a risk if we refactor things. I'm not sure if there's a better way though and maybe we should just accept that.

ekohl (commented):
At which point do these become big enough to deserve a proper package? Is this something that at some point will migrate to foreman-maintain?

mbacovsky:
@ekohl there are plans to move backup and restore to the foreman-maintain and I believe it is a matter of the resources now. Is it @ehelms? 

In foreman-maintain some time ago I started the work on extending the `features` representing foreman and candlepin databases to support remote DBs. I used the service configs to obtain the db connection data so it is not necessary to couple it that tightly with puppet modules (answer files). Not sure which approach is better.

ehelms:
That is the general plan, although no one currently owns migrating them to foreman-maintain. But we have talked about it, and proposed it via an RFC. I don't think it would be a ton of work, more work on the testing side, and updating docs to incorporate using foreman-maintain to perform the actions versus these scripts.

ehelms:

johnpmitsch:
This functionality is being added to foreman maintain and here: https://github.com/Katello/katello-installer/pull/550

PR #501, opened 2017-08-17
Fixes #20632 - require pg_globals for katello-restore

johnpmitsch:


cfouant:
I noticed the `backup_invalid_message` states that there are too many files in the backup directory, and if the file is missing it's actually because there aren't enough files in the backup directory. After looking at the `display_backup_options` I think we should get rid of `backup_invalid_message` and simply go right into `display_backup_options`.

johnpmitsch:
@cfouant +1, updated

cfouant:
ACK thanks @johnpmitsch !

PR #505, opened 2017-08-18
fixes #20650 - validates postgres has write access to backup directory

cfouant:


johnpmitsch:
@cfouant Something about restricting backups to /tmp and /var/tmp doesn't sit well with me, I have a few suggestions about alternatives, curious to hear your thoughts:
1) Can we just chown the backup directory to postgres? We do that in [satellite-clone](https://git.io/v5e3M)
2) Could we check if postgres has access to the folder and fail with an error message if not? It would look something like this (would have to rubify this)
```
[vagrant@centos7-katello-3-4 ~]$ sudo -u postgres test -r /backup/
[vagrant@centos7-katello-3-4 ~]$ echo $?
0
```
any non-zero exit code would mean postgres user doesn't have read access to the folder.

cfouant:
@johnpmitsch actually we can chown the backup directory to postgres, which we do. But if the base directory doesn't have write access for postgres, then Postgres still cannot write to it. And chowning all the way down to root would be a bad thing.

It also states in the docs that we can only write to /var/tmp or /tmp - this simply makes a failure easier to pinpoint why.

I was looking at your version (checking if postgres has write access) but wasn't really finding any neat ways to do it with ruby. I'll keep looking and see what I can find, but as it stands, we still have the docs saying 'only /tmp or /var/tmp', so this matches up with the docs.

stbenjam:
Why not backup only postgres to /var/tmp and then move it to the target directory?

johnpmitsch:
@cfouant Let me share an example of what I am thinking:

This is recreating the steps we take [here](https://git.io/v5TiT) for two different directories, one in `/` and the other in `/root/`
```
irb(main):001:0> has_access = "/has-access"
=> "/has-access"
irb(main):002:0> no_access = "/root/no-access/"
=> "/root/no-access/"
irb(main):004:0> require "fileutils"
=> true
irb(main):005:0> FileUtils.mkdir_p has_access
=> ["/has-access"]
irb(main):006:0> FileUtils.chown_R nil, 'postgres', has_access
=> ["/has-access"]
irb(main):007:0> FileUtils.chmod_R 0770, has_access
=> ["/has-access"]
irb(main):008:0> FileUtils.mkdir_p no_access
=> ["/root/no-access/"]
irb(main):009:0> FileUtils.chown_R nil, 'postgres', no_access
=> ["/root/no-access/"]
irb(main):010:0> FileUtils.chmod_R 0770, no_access
=> ["/root/no-access/"]
```

Here is the part I think we should add, its testing if postgres has write access to that directory
```
irb(main):013:0>  `sudo -u postgres test -w  #{has_access}`
=> ""
irb(main):014:0> $?.success?
=> true
irb(main):015:0>  `sudo -u postgres test -w  #{no_access}`
=> ""
irb(main):016:0> $?.success?
=> false
```

The directory in `/root` didn't have write access because its parent directory (/root) is owned by root, and we were able to see that using `$?.success?` of the test command. We can do the same in the script and print out an error message to the user if the test command fails.

I'm in favor of doing a check like this and leaving it up to the user to ensure the backup directory is in the correct location, this will ensure the psql dumps will run without permission errors and allow the users to place the backup directory wherever it will work.

ehelms:
@cfouant This looks to have stagnated, what can we do to get this finished up?

cfouant:
@ehelms - it'll be in before tomorrow! Just tweaking a few details.

katello/katello-backup
```diff
@@ -355,6 +355,20 @@ def backup_config_files
...
+  unless postgres_write_access?
+    puts "****cancelled****"
+    puts "Postgres user needs write access to the backup directory"
+    puts "Please select a directory, such as /tmp or /var/tmp which allows Postgres write access"
+    cleanup
+  end
+end
+
+def postgres_write_access?
+  access = `sudo -u postgres test -w #{@dir} | echo $?`.strip
```
jlsherrill: using | to run both doesn't seem to work properly, and it'd be better to use system() to get the return status:

`system("sudo -u postgres test -w #{@dir}")`

this returns true if exit code is zero (meaning it could write), returns false if not

jlsherrill (approved):
ACK

jlsherrill (approved):

johnpmitsch (approved):
Works well, nice change!
error message and cleanup on directory w/o permissions
```
[root@change-hostname-34 katello]# ./katello-backup /root/
Starting backup: 2017-09-29 16:06:30 +0000
Creating backup folder /root/katello-backup-20170929160630
****cancelled****
Postgres user needs write access to the backup directory
Please select a directory, such as /tmp or /var/tmp which allows Postgres write access
Cleaning up backup folder and starting any stopped services... 
```
and runs successfully on directories expected to work
```
[root@change-hostname-34 katello]# ./katello-backup /backup/
Starting backup: 2017-09-29 16:08:59 +0000
Creating backup folder /backup/katello-backup-20170929160900
Generating metadata ... 
Done.
Backing up config files... 
Done.
```

BenGig:
Checking the backup directory for postgres access makes sense with online backup mode only. Offline backup should skip the test because the postgres user is not involved.

PR #507, opened 2017-08-23
Fixes #20702 - require python-argparse on EL5 and EL6

evgeni:
python-argparse is available in EPEL5 for EL5 and in EL6 proper


evgeni:
nope, let's just use optparse

PR #511, opened 2017-08-28
Fixes #20768 - kchange-hostname use variables for names

johnpmitsch:


johnpmitsch:
Going with a different solution for this

PR #517, opened 2017-08-30
Fixes #20805 - kchange-hostname fix optparser error

johnpmitsch:
optparser isn't recognized inside of functions, so its changed to an instance variable here, and then the prechecks are changed to run before the warning message


johnpmitsch:
To test

spin up a katello 3.4 production box and run:
```
yum install -y git vim
cd ~
cd katello-packaging
git fetch fork
git checkout fork/opt_parser_bug
cd ~
```

try excluding options and make sure you get an error message and printout of the options. i.e.
`katello-change-hostname  newname` (exclude -u and -p options)

then make sure proper input works as well, test any other edge cases you can think of

jturel (approved):
Working well for me. Nice improvement!

PR #519, opened 2017-08-30
Service improvements

ekohl:
See the individual commits for details.

### DIFF
--- a/katello/katello.spec
+++ b/katello/katello.spec
@@ -174,6 +174,7 @@ Requires:       wget
 Requires:       curl
 Requires:       ruby
 Requires:       net-tools
+Requires:       /bin/systemctl
 
 %description service
 Useful utilities for managing Katello services
--- a/katello/katello.spec
+++ b/katello/katello.spec
@@ -173,6 +173,7 @@ Group: Applications/System
 Requires:       wget
 Requires:       curl
 Requires:       ruby
+Requires:       net-tools
 
 %description service
 Useful utilities for managing Katello services
--- a/katello/service-wait
+++ b/katello/service-wait
@@ -91,7 +91,7 @@ after_start() {
 before_stop() {
   case "$SERVICE" in
     tomcat6|tomcat)
-      # RHBZ 789288 - wait until service port is avaiable
+      # RHBZ 789288 - wait until service port is available
       for i in {1..$WAIT_MAX}; do netstat -ln | grep -q ":$TOMCAT_SERV_PORT\s" && break; sleep 1; done
       ;;
   esac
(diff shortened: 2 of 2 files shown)


ekohl:
Based on https://github.com/Katello/puppet-service_wait/pull/28#pullrequestreview-59552590. I opted for the simple solution and just depend on net-tools. This whole PR might be a bit of a request for feedback.

katello/katello.spec
```diff
@@ -173,6 +173,7 @@ Group: Applications/System
 Requires:       wget
 Requires:       curl
 Requires:       ruby
+Requires:       net-tools
```
sean797: What about `Requires: /bin/systemctl` ?

sean797 (changes requested):

katello/katello.spec
ekohl: Good one. I also considered depending on ```/usr/bin/netstat``` rather than ```net-tools``` or rewriting it to just depend on ```/usr/bin/ss```

sean797: Im not sure that will work:

```
[root@ssean ~]$ rpm -qf /usr/bin/netstat
net-tools-2.0-0.42.20160912git.fc26.x86_64
[root@ssean ~]$ rpm -q --provides net-tools-2.0-0.42.20160912git.fc26.x86_64
net-tools = 2.0-0.42.20160912git.fc26
net-tools(x86-64) = 2.0-0.42.20160912git.fc26
```

Whereas systemctl:

```
[root@ssean ~]$ rpm -qf /usr/bin/systemctl
systemd-233-6.fc26.x86_64
[root@ssean ~]$ rpm -q --provides systemd-233-6.fc26.x86_64
/bin/systemctl
/sbin/shutdown
config(systemd) = 233-6.fc26
libsystemd-shared-233.so()(64bit)
pkgconfig(systemd) = 233
pkgconfig(udev) = 233
syslog
system-setup-keyboard = 0.9
systemd = 233-6.fc26
systemd(x86-64) = 233-6.fc26
systemd-sysv = 206
systemd-units = 233-6.fc26

sean797: So i'd say stick with `net-tools` unless you want to re-write it for `ss`.

sean797 (approved):
Looks good! Thanks @ekohl 👍

jlsherrill (approved):
thanks @ekohl

katello/katello.spec
ekohl: Sounds like rewriting to ```ss``` is better in the long run where Fedora is removing net-tools https://pagure.io/fesco/issue/1711

PR #520, opened 2017-08-30
Release katello-client-bootstrap 1.4.1

evgeni:


katello-client-bootstrap/katello-client-bootstrap.spec
```diff
@@ -49,6 +49,9 @@ cp bootstrap.py %{buildroot}%{_var}/www/html/pub/bootstrap.py
 %{_var}/www/html/pub/bootstrap.py
 
 %changelog
+* Wed Aug 30 2017 Evgeni Golov <egolov@redhat.com> 1.4.1-1
+- Release katello-client-bootstrap 1.4.1 (egolov@redhat.com)
+
```
ehelms: Tito should make this entry for us when we release it. Not sure what it will do if it finds this in existence already. But I can merge and try.

evgeni: I think (looking at 1.4.0) it just adds another one. But I can also yank that?

ehelms (approved):

PR #521, opened 2017-08-31
python2-tracer 0.6.12 only in EPEL and F26

ehelms:
F25 lacks the high enough version to support tracer:  http://linux.mirrors.es.net/fedora/releases/25/Everything/x86_64/os/Packages/p/python2-tracer-0.6.11-1.fc25.noarch.rpm

### DIFF
--- a/katello-host-tools/katello-host-tools.spec
+++ b/katello-host-tools/katello-host-tools.spec
@@ -13,7 +13,7 @@ BuildArch: noarch
 Requires: subscription-manager
 Requires: %{name}-fact-plugin
 
-%if 0%{?fedora} > 18 || 0%{?rhel} > 6
+%if 0%{?fedora} > 24 || 0%{?rhel} > 6
 Requires: python2-tracer >= 0.6.12
 Requires: python-rhsm
 Requires: crontabs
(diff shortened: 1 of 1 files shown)


ehelms:
@sean797 Seem reasonable to you or can we relax the tracer version requirement?

sean797 (approved):
Unfortunately we need a minimum of that version for https://github.com/FrostyX/tracer/commit/d2f5164a4246b95798a35617dbd66cbc1f0c68b1 `.now()` (used in https://github.com/Katello/katello-agent/blob/master/src/yum-plugins/tracer_upload.py#L36)

Weird I can't see it in the repos but https://bodhi.fedoraproject.org/updates/FEDORA-2016-3f9b4e3c43 shows it should have been pushed to stable.

ehelms:
@sean797 Ahhh, its in the updates repository (now the Everything repo) [1]. I will just add that to repoclosure for F25 which should mean this check is no longer needed. What I would ask if you would be willing to update the katello client docs to include this requirement for F25?


[1] http://ftp.linux.ncsu.edu/pub/fedora/linux/updates/25/x86_64/p/

sean797:
Sure, but I'm not sure they need updating? The updates repo should be enabled on all Fedora installs by default.

Do you want to update you're change to  `>  22` since its only available on F23+.

ehelms:
@sean797 Updated

katello-host-tools/katello-host-tools.spec
```diff
@@ -161,7 +161,7 @@ exit 0
 %attr(750, root, root) %{_sbindir}/katello-enabled-repos-upload
 %{_prefix}/lib/yum-plugins
 
-%if 0%{?fedora} > 18 || 0%{?rhel} > 6
+%if 0%{?fedora} > 25 || 0%{?rhel} > 6
```
sean797, line 164: Typo ? The other lines are `24`

sean797 (changes requested):

ehelms:

komidore64:
[test katello-packaging]

komidore64:
[test katello-packaging]

PR #522, opened 2017-09-01
Fixes #20830 - add man page for katello-change-hostname

johnpmitsch:


katello/katello.spec
```diff
@@ -26,6 +26,7 @@ Source9:    qpid-core-dump
 Source10:   katello-clean-empty-puppet-environments
 Source11:   katello-change-hostname
 Source12:   katello-repository-publish-check
+Source13:   katello-change-hostname.8.asciidoc
```
ehelms, line 29: This only adds it as a source and won't ensure its deployed properly to the system.

johnpmitsch: @ehelms updated the man page

johnpmitsch:
After installing katlello and katello-common, the man page works as expected:
```
[root@foo ~]# man katello-change-hostname 
[root@foo ~]# echo $?
0
```

ehelms (approved):

PR #523, opened 2017-09-05
Installation and upgrade times and catalina.out

lzap:
This adds key installation/upgrade logs from yum.log and also fixes `catalina.out` (note the different extension there).

### DIFF
--- a/katello/katello-debug.sh
+++ b/katello/katello-debug.sh
@@ -33,6 +33,9 @@ add_files /etc/foreman-installer/*
 add_cmd "find /root/ssl-build -ls | sort -k 11" "katello_ssl_build_dir"
 add_cmd "find /etc/pki -ls | sort -k 11" "katello_pki_dir"
 
+# Installation and upgrade times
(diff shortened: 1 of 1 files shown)


katello/katello-debug.sh
```diff
@@ -33,6 +33,9 @@ add_files /etc/foreman-installer/*
 add_cmd "find /root/ssl-build -ls | sort -k 11" "katello_ssl_build_dir"
 add_cmd "find /etc/pki -ls | sort -k 11" "katello_pki_dir"
 
+# Installation and upgrade times
+add_cmd "egrep '(katello|satellite)' /var/log/yum.log" "yum_install_upgrade_log"
```
sean797, line 37: Should `satellite` be a downstream only change?

ehelms: Do we have this for Foreman packages? and plugins?

lzap: I am fine removing "satellite" here. @ehelms not sure I get what you mean tho?

ehelms:

ehelms:

PR #524, opened 2017-09-05
Fixes #20821 - Separate katello-* utilities into classes

johnpmitsch:
This moves katello-backup, restore, and hostname change into classes in lib/ that will be called by the actual scripts. This allows us to have more flexibility in changing terminology in backports and also makes the script functionality more testable and compartmentalized.

TO-DO:

- [x] update spec files


johnpmitsch:
To test, spin up a 3.4 box and run:
```
mkdir /usr/share/katello
cd ~
cd katello-packaging/
git fetch --all
git checkout fork/katello_script_refactor
ln -s /root/katello-packaging/katello/lib/restore.rb /usr/share/katello/restore.rb
ln -s /root/katello-packaging/katello/lib/backup.rb /usr/share/katello/backup.rb
ln -s /root/katello-packaging/katello/lib/hostname-change.rb /usr/share/katello/hostname-change.rb
ln -s /root/katello-packaging/katello/lib/script-helper.rb /usr/share/katello/script-helper.rb
cd ~
ln -s /root/katello-packaging/katello/katello-change-hostname katello-change-hostname
ln -s /root/katello-packaging/katello/katello-restore katello-restore
ln -s /root/katello-packaging/katello/katello-backup katello-backup
chmod a+x katello-restore
chmod a+x katello-backup
```
All backup/restore/change-hostname functionality should be the same as usual

katello/lib/backup.rb
```diff
@@ -0,0 +1,420 @@
...
+require 'find'
+require 'json'
+require 'highline/import'
+require "/usr/share/katello/script-helper.rb"
+
+class Backup
+  include ScriptHelper
+
+  def initialize(foreman_proxy_content, foreman_proxy)
+    @excluded="--exclude goferd,foreman-proxy,squid,smart_proxy_dynflow_core,qdrouterd,qpidd"
```
sean797: I know you haven't changed this but doesn't qpidd have to stop to get a consistent backup of `/var/lib/qpidd` ?

sean797 (commented):
Haven't tested this but just looking over it I think changes to the `.spec` file will also be need

katello/lib/backup.rb
johnpmitsch: Thanks @sean797, I marked it WIP as I do still have that as a change to make, but wanted to get eyes on this early. 

I'm not sure about qpidd  stopping to get an accurate backup, maybe @jlsherrill or @ehelms can comment

jturel:
Just starting my own review but I'm curious if you see any value in one or more of these options:

- Putting the Backup script into module => Katello::Backup.new
- Renaming Backup to BackupCreator which I think makes the intent of the class more clear

Ditto on those suggestions for the other scripts you adjusted.

johnpmitsch:
@jturel I am for having a namespace to make things clear, but one thing I do want to avoid is having a useless subdirectory, i.e. `katello/lib/katello/backup.rb` vs `katello/lib/backup.rb`. I think I can add the modules without the subdirectory (I don't think this breaks conventions or best practices), I'll do that.

Let me think about the names a little more, one idea I have is this:
```
Tools::Backup
Tools::Restore
Tools::HostnameChange
```
or 
```
Utilities::Backup
Utilities::Restore
Utilities::HostnameChange
```
That would signify that these are tools and not actual backups, etc.
I'm not sure if it needs a specific Katello declaration (since we aren't in a plugin situation like we are in actual Katello repo), but open to it if others agree.
\</bikeshedding>

katello/katello-backup
```diff
@@ -1,405 +1,10 @@
...
+require "/usr/share/katello/backup.rb"
 
-require 'optparse'
-require 'fileutils'
-require 'date'
-require 'yaml'
-require 'find'
-require 'json'
-require 'highline/import'
+katello_backup = Utilities::Backup.new("foreman-proxy-content", "foreman proxy")
```
ehelms: Utilities is a bit of a broad namespace, I do think you'd be better off with something Katello or Foreman specific.

katello/katello.spec
```diff
@@ -85,6 +85,13 @@ install -m 755 %{SOURCE3} %{buildroot}%{_sysconfdir}/cron.weekly/katello-remove-
 install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
 install -m 755 %{SOURCE12} %{buildroot}%{_sysconfdir}/cron.daily/katello-repository-publish-check
 
+# symlink script libraries
+mkdir -p %{_datarootdir}/katello
+ln -s lib/restore.rb %{_datarootdir}/katello/restore.rb
+ln -s lib/backup.rb %{_datarootdir}/katello/backup.rb
+ln -s lib/hostname-change.rb %{_datarootdir}/katello/hostname-change.rb
+ln -s lib/script-helper.rb %{_datarootdir}/katello/script-helper.rb
```
ehelms: All of these new files will need to be included as Source at the top of the spec.

katello/utilities/helper.rb
```diff
@@ -0,0 +1,15 @@
...
+    end
+
+    def accepted_scenarios
+      ["katello", "foreman-proxy-content"]
+    end
+
+    def error_message
+      "This utility can't run on a non-katello system."
+    end
+  end
```
ehelms: This helper script will make some parts of re-using these scripts more difficult. Or rather, not simply straightforward.

katello/katello.spec
ehelms: Instead of symlinks, you can see down below, starting at line 98, you can just have them installed to the location you expect them at.

ehelms: Take a look at line 135 -- you'll need to include all of these so they get laid down on disk.

katello/katello-backup
johnpmitsch: KatelloUtilities seems like it would be pretty explicit about what it contains. I'll switch to that unless there are objections

jturel:
I was able to test each of the scripts and see that they are working as intended! I'll accept once the module is renamed. Deferring to Eric on the spec file changes (though, they make sense).

ehelms:
@johnpmitsch What's the plan here? I see a lot of fixes going into some of these scripts thus causing this to need a rebase and update. Which will lead this to need to be re-tested.

johnpmitsch:
@ehelms I have to figure that out, its probably going to be that I manually apply all of the recent commits to the katello-change hostname class

johnpmitsch:
I updated this with the latest katello-change-hostname commits and renamed the `Utitlies` class to `KatelloUtilities`. 
@evgeni  Can you make sure all of the latest commits to katello-change-hostname are found in here? I manually applied them since I couldn't rebase properly with the refactoring.
@ehelms Can you check the spec file?
@jturel Can you retest? Here is what an updated version of what I use to test
```
yum install -y git vim
mkdir /usr/share/katello
cd ~
cd katello-packaging/
git fetch --all
git checkout fork/katello_script_refactor
ln -s /root/katello-packaging/katello/utilities/restore.rb /usr/share/katello/restore.rb
ln -s /root/katello-packaging/katello/utilities/backup.rb /usr/share/katello/backup.rb
ln -s /root/katello-packaging/katello/utilities/hostname-change.rb /usr/share/katello/hostname-change.rb
ln -s /root/katello-packaging/katello/utilities/helper.rb /usr/share/katello/helper.rb
cd ~
ln -s /root/katello-packaging/katello/katello-change-hostname katello-change-hostname
ln -s /root/katello-packaging/katello/katello-restore katello-restore
ln -s /root/katello-packaging/katello/katello-backup katello-backup
chmod a+x katello-restore
chmod a+x katello-backup
```

jturel:
I retested all scripts and they are working well.

katello/katello.spec
```diff
@@ -26,6 +26,10 @@ Source9:    qpid-core-dump
 Source10:   katello-clean-empty-puppet-environments
 Source11:   katello-change-hostname
 Source12:   katello-repository-publish-check
+Source13:   utilities/restore.rb
+Source14:   utilities/backup.rb
+Source15:   utilities/hostname-change.rb
+Source16:   utilities/helper.rb
```
mbacovsky: @johnpmitsch  I'm afraid it is not possible to have resources in subdirectories [1]. It needs to go to the root dir or in a tarball. 

What were the reasons originally to put these scripts here and not in the katello repo?
Was there any agreement on plan to move these utilities to `foreman-maintain`?

[1] http://rikers.org/rpmbook/node68.html#SECTION03526100000000000000

johnpmitsch: @mbacovsky The main reason of this refactoring is to make it easy to change our naming conventions for backporting these scripts to satellite. But this is also going to make the scripts more testable and is the first step to moving them to foreman-maintain.

I'm not sure what do do about the spec file and handling the subdirectory, if anyone who is more experienced with rpm spec files can help me out, that would be appreciated

johnpmitsch:
After some off-thread IRC conversations, I updated the scripts to live in the `katello/` dir, everything. I'm currently re-testing the scripts and so far, so good.

@jturel apologies, but do you mind retesting? You will just have to update the links in the setup script that I put in a previous comment. (thanks for being patient)

@ehelms I'm not sure what is going wrong with the build in jenkins, but I suspect something is off in my spec file. Can you take a look? I can't seem to get the exact error (The output mentions checking build.log).

katello/katello.spec
```diff
@@ -85,6 +89,13 @@ install -m 755 %{SOURCE3} %{buildroot}%{_sysconfdir}/cron.weekly/katello-remove-
 install -m 755 %{SOURCE10} %{buildroot}%{_sysconfdir}/cron.weekly/katello-clean-empty-puppet-environments
 install -m 755 %{SOURCE12} %{buildroot}%{_sysconfdir}/cron.daily/katello-repository-publish-check
 
+# symlink script libraries
+mkdir -p %{_datarootdir}/katello
```
mbacovsky: It seems you need to prefix these with `%{buildroot}` see the build.log http://koji.katello.org/kojifiles/work/tasks/9917/29917/build.log

johnpmitsch: Thanks! I updated

mbacovsky:
The build is green now, the rpm can be downloaded from the task [1] for easier testing.
[1] http://koji.katello.org/koji/taskinfo?taskID=29919

jturel:
Everything is still good with the change to the KatelloUtilities namespace. 👍

johnpmitsch:
@jturel @evgeni Can you double check that all the changes that have been made since this PR has been open are correctly included in this PR? I wasn't able to resolve any merge conflicts since the files moved and changed so drastically, so applied them manually.

https://github.com/Katello/katello-packaging/commits/master/katello/katello-change-hostname

johnpmitsch:
Updating to use accept an argument for the accepted_scenarios (again, for backporting). Changes are in a separate commit

This will allow us to overwrite the `accepted_scenarios` method, as shown below:
```
module Foo
  def scenarios
    @accepted || ["in module"]
  end
end

class Bar
  include Foo

  def initialize(accepted=nil)
    @accepted = accepted
  end
end
```
```
2.2.4 :002 > bar_no_params = Bar.new
 => #<Bar:0x00000000b1dcd0 @accepted=nil> 
2.2.4 :003 > bar_no_params.scenarios
 => ["in module"] 
2.2.4 :004 > bar_with_params = Bar.new(["not in module"])
 => #<Bar:0x00000000a8cb90 @accepted=["not in module"]> 
2.2.4 :005 > bar_with_params.scenarios
 => ["not in module"] 
```

johnpmitsch:
Testing out the accepted scenarios changing:
```diff
diff --git a/katello/katello-backup b/katello/katello-backup
index 745143e..f74b3c8 100755
--- a/katello/katello-backup
+++ b/katello/katello-backup
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 require "/usr/share/katello/backup.rb"
 
-katello_backup = KatelloUtilities::Backup.new("foreman-proxy-content", "foreman proxy")
+katello_backup = KatelloUtilities::Backup.new("foreman-proxy-content", "foreman proxy", ["blah"])
 
 if katello_backup.accepted_scenarios.include? katello_backup.last_scenario
   katello_backup.run
diff --git a/katello/katello-change-hostname b/katello/katello-change-hostname
index b5029c0..e4d5f8d 100755
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 require "/usr/share/katello/hostname-change.rb"
 
-hostname_change = KatelloUtilities::HostnameChange.new("Foreman Proxy", "Foreman Proxies", "foreman-proxy-content")
+hostname_change = KatelloUtilities::HostnameChange.new("Foreman Proxy", "Foreman Proxies", "foreman-proxy-content", nil, nil, ["blah"])
 
 if hostname_change.accepted_scenarios.include? hostname_change.last_scenario
   hostname_change.run
diff --git a/katello/katello-restore b/katello/katello-restore
index e0f7bca..c3fce8f 100755
--- a/katello/katello-restore
+++ b/katello/katello-restore
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 require "/usr/share/katello/restore.rb"
 
-katello_restore = KatelloUtilities::Restore.new("foreman-proxy-content")
+katello_restore = KatelloUtilities::Restore.new("foreman-proxy-content", ["blah"])
 
 if katello_restore.accepted_scenarios.include? katello_restore.last_scenario
   katello_restore.run
```
```
[root@change-hostname-34 katello]# ./katello-backup /backup
This utility can't run on a non-katello system.
[root@change-hostname-34 katello]# ./katello-restore /backup
This utility can't run on a non-katello system.
[root@change-hostname-34 katello]# ./katello-change-hostname -u admin -p changeme -d foo.example.com
This utility can't run on a non-katello system.
```
An error message is displayed when you pass in a parameter that is not an accepted scenario to the classes

jturel:
@johnpmitsch All the porting of recent changes (from 9-14 to present) look good!

johnpmitsch:
@ehelms can you look over the recent changes?

katello/backup.rb
```diff
@@ -0,0 +1,423 @@
+require 'optparse'
+require 'fileutils'
+require 'date'
+require 'yaml'
+require 'find'
+require 'json'
+require 'highline/import'
+require "/usr/share/katello/helper.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

How about

```ruby
require File.join(File.expand_path('..', __FILE__), 'helper.rb')
```

instead? :)

katello/backup.rb
```diff
@@ -0,0 +1,423 @@
...
+      @accepted_scenarios = accepted_scenarios
+
+      # keep as variables for easy backporting
+      @foreman_proxy_content = foreman_proxy_content
+      @foreman_proxy = foreman_proxy
+
+      setup_opt_parser
+    end
+
+    def run_cmd(command, exit_codes=[0])
```
evgeni: Should this be moved to `helper.rb`?

katello/hostname-change.rb
```diff
@@ -0,0 +1,363 @@
...
+      @options = {}
+      @options[:program] = program || @default_program
+      @options[:scenario] = scenario || @last_scenario
+      @options[:system_check] = false
+      @foreman_proxy_content = @options[:scenario] == @proxy_hyphenated
+
+      setup_opt_parser
+    end
+
+    def run_cmd(command, exit_codes=[0], message=nil)
```
evgeni: Should this be moved to `helper.rb`?

katello/hostname-change.rb
```diff
@@ -0,0 +1,363 @@
...
+      when "katello"
+        return "foreman"
+      when @proxy_hyphenated
+        return "foreman"
+      else
+        return @last_scenario
+      end
+    end
+
+    def timestamp
```
evgeni: Should this be moved to `helper.rb`?

katello/hostname-change.rb
```diff
@@ -0,0 +1,363 @@
...
+      Time.now.strftime("%Y%m%d%H%M")
+    end
+
+    def get_hostname
```
evgeni: Should this be moved to `helper.rb`?

katello/hostname-change.rb
```diff
@@ -0,0 +1,363 @@
...
+        return true
+      elsif str.chr.downcase == "n"
+        return false
+      else
+        puts "Invalid Character. Try again: [y/n]"
+        self.yesno
+      end
+    end
+
+    def fail_with_message(message, opt_parser=nil)
```
evgeni: Should this be moved to `helper.rb`?

katello/restore.rb
```diff
@@ -0,0 +1,275 @@
+require 'socket'
+require 'optparse'
+require 'fileutils'
+require 'date'
+require "/usr/share/katello/helper.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

katello/katello-restore
```diff
@@ -1,257 +1,10 @@
 #!/usr/bin/env ruby
+require "/usr/share/katello/restore.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

katello/katello-backup
```diff
@@ -1,405 +1,10 @@
 #!/usr/bin/env ruby
+require "/usr/share/katello/backup.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

katello/katello-change-hostname
```diff
@@ -1,350 +1,10 @@
-#!/usr/bin/ruby
+#!/usr/bin/env ruby
+require "/usr/share/katello/hostname-change.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

katello/hostname-change.rb
```diff
@@ -0,0 +1,363 @@
+require "socket"
+require "optparse"
+require "rubygems"
+require "yaml"
+require "shellwords"
+require "json"
+require "uri"
+require "/usr/share/katello/helper.rb"
```
evgeni: This makes it impossible to test the code from a git checkout :(

katello/katello.spec
```diff
@@ -26,6 +26,10 @@ Source9:    qpid-core-dump
 Source10:   katello-clean-empty-puppet-environments
 Source11:   katello-change-hostname
 Source12:   katello-repository-publish-check
+Source13:   restore.rb
```
evgeni: at this point I wonder if we should create a proper katello_utils gem, put all the real code there, and just package require that?

evgeni (changes requested):
after reading this, I think we should go a step further and make the code a real gem, not a random collection of .rb files that live in katello-packaging.

I did not verify if you can or cannot do copy&paste properly ;)

katello/katello.spec
mbacovsky: @evgeni I thought there was general agreement on moving the utils to `foreman-maintain`in a near future. Do you think it makes sense to invest time in creating a temporary gem?

evgeni: Right, I remember hearing that. I guess it highly depends on the time-schedule then?
Creating a gem is (IMHO) pretty easy and cheap, unless we're moving to foreman-maintain in 2 wks or something.
The code requires a class-ification for either path (pure class-ification, a k_u gem or f_m gem) and if make a gem now, the move to foreman-maintain is just a rename of the files away.

katello/backup.rb
johnpmitsch: @evgeni I put [instructions](https://git.io/vdklj) here a few times on how to check this out for testing. Those will be a little out of date, but I also tried to update it in newer comments (that may be hidden by github). 

Will the above solution place the files in `/usr/share/katello` in a production install? That is where we agreed to put the utility class files

evgeni: The above does not move the files anywhere, it just requires them with a relative path.
We even could just use require_relative here in most cases.

johnpmitsch:
@evgeni This is meant to be a "step in the right direction" change to start the utilities moving over to foreman-maintain. I am not sure I would want to change this to be a gem before doing that, I plan to start on the effort after this gets merged.

I agree on the further DRYing up of the classes, but wanted to keep things simple in this first refactoring and then follow up with smaller additional changes (moving other scripts into classes and removing duplicate commands). I would rather merge this first so any additional changes anyone makes to the three refactored scripts can use this structure and I don't have to keep manually applying changes to master to this PR. Of course, as always, I'm open to discussion :)

katello/backup.rb
johnpmitsch: @evgeni how does that work with the rpm install?

katello/katello.spec
mbacovsky: @evgeni I see your point and it makes sense. However I don't expect the move to be just a rename. `foreman-maintain` is supposed to work on all the scenarios we have (foreman, katello, various combination of the plugins, upstream, downstream). f-m as a framework introduces `features`and behaves on their presence or absence. Some of the logic is already there some will need to be extended. I expect the code here to get restructured to match f-m's layout. With that in mind I just wanted to rise discussion if further classification still makes sense.

johnpmitsch:
@evgeni I updated according to your suggestions, can you give it a second look?

ehelms:
@johnpmitsch I merged your man-page PR and I think I broke the spec file. Please rebase.

ehelms:
@evgeni As stated, the general agreement and discussion was to do this a as a step towards foreman_maintain, without going full scale in creating a new gem, new packaging, new everything. While this will likely persist a full release, I don't agree with breaking it out at this time. Yes, it makes this place uglier, but the intent was to do as little as possible to re-factor into re-usable classes to facilitate easier re-use and a shortened timescale. Another way to look at it is, we want to fix the "bug" not turn this into a feature.

I should also note that I agree with @johnpmitsch in that, in this case, re-factoring here will lead to higher risk rather than lower risk due to the amount and complexity of changes that are introduced. By first wrapping existing functionality into classes with as few new methods as possible we retain the original behaviors enough to regression test easier. The more that changes, the more edge cases we potentially introduce on untested code.

The ideal move to foreman_maintain will introduce rubocop, and testing and a singular framework for use which will provide a lot more robustness.

ehelms:
Looks like yet again a round of re-factors has occurred meaning this will need the full gambit of re-testing. Does that fall to you @jturel ?

johnpmitsch:
@ehelms thanks, I rebased the man page, please double check it looks ok. @evgeni is going to take another look at the latest refactor

@jturel I would recommend holding off on testing until @evgeni takes a look and comments, that way we can save you some unnecessary testing if there are additional changes :)

katello/restore.rb
```diff
@@ -0,0 +1,266 @@
...
+          @disable_system_checks = true
+        end
+      end
+      @optparse
+    end
+
+    def parse_options
+      begin @optparse.parse! ARGV
+        if ARGV.length == 0
+          opts.abort("**** ERROR: Please specify the backup directory to restore ****")
```
evgeni: this has to be `@optparse`, not `opts`

katello/restore.rb
```diff
@@ -0,0 +1,266 @@
...
+        elsif ARGV.length != 1
+          puts opts
```
evgeni: this has to be `@optparse`, not `opts`

katello/restore.rb
```diff
@@ -0,0 +1,266 @@
...
+          exit(-1)
+        end
+
+        @dir = ARGV.pop.dup
+        unless File.directory?(@dir)
+          opts.abort("Backup directory does not exist: #{@dir}")
```
evgeni: this has to be `@optparse`, not `opts`

evgeni (commented):
This is clearly a step in the right direction! I think going without the gem until we merge into foreman-maintain is OK - esp as I missed the "we need to do the packaging for the gem" part in my initial "it's soo easy" comment.

There is one comment on restore.rb. ACK as soon as that has been adressed!

:sheep:

johnpmitsch:
@evgeni nice catch! I updated @jturel lets test this out again (hopefully for the last time)

ehelms (approved):

jturel:
I was able to verify all scripts against a 3.4 installation and a proxy!

PR #526, opened 2017-09-14
Fixes #20919 - use Shellwords.escape to escape hammer user/pass

evgeni:


johnpmitsch (approved):
awesome hammer_cmd wrapper!

PR #527, opened 2017-09-14
Fixes #20921 - use String.chomp instead of slicing to remove newlines

evgeni:


johnpmitsch (approved):
```
irb(main):005:0> `hammer -u admin -p changeme$ capsule list | grep "changed34" | awk '{ print $1 }'`.chomp
=> "1"
```
Much better way of doing things, thanks! :)

johnpmitsch:
@evgeni just needs a rebase from the `hammer_cmd` PR

evgeni:
rebased

PR #528, opened 2017-09-14
Fixes #20922 - use highline for user promting

evgeni:


evgeni:
rebased

sean797 (approved):
ACK. Thanks @evgeni 👍 Sorry about the late review

ehelms:
highline is not default ruby is it? So the spec files that include this file will need to require this package now?

evgeni:
@ehelms you are right, even though katello-backup was using highline before, w/o explicitly adding that dep. added now.

evgeni:
[test]

ehelms:

evgeni:
moved to https://github.com/theforeman/foreman-packaging/pull/1928

PR #529, opened 2017-09-14
Fixes #20924 - check credentials before doing any changes

evgeni:
depends on https://github.com/Katello/katello-packaging/pull/526

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -171,6 +171,12 @@ def precheck
     fail_with_message("There is a problem with the server, please check 'hammer ping'")
   end
 
+  STDOUT.puts "\nChecking credentials"
+  hammer_cmd("capsule list")
+  unless $?.success?
+    fail_with_message("There is a problem with the credentials, please retry")
+  end
+
   if @options[:confirm]
     response = true
   else


katello/katello-change-hostname
```diff
@@ -170,6 +171,12 @@ def precheck
     fail_with_message("There is a problem with the server, please check 'hammer ping'")
   end
 
+  STDOUT.puts "\nChecking credentials"
```
johnpmitsch, line 174: I was trying to think if there is a more generic command we can use, but can't really find one! Maybe @akofink can think of one (we are just trying to check if the username/password credentials are correct)

katello/katello-change-hostname
```diff
@@ -170,6 +171,12 @@ def precheck
...
+  hammer_cmd("capsule list")
+  unless $?.exitstatus == 0
```
johnpmitsch: can we use `$?.success?` syntax here? I use it in other places in the script.

johnpmitsch (changes requested):
just a couple small comments

katello/katello-change-hostname
akofink: `hammer ping` ?

evgeni: `hammer ping` does not require credentials, and it does not error out for wrong ones either.
we could do `hammer user info --name #{@options[:username]}`, but that feels equally ugly.

katello/katello-change-hostname
evgeni: Yes we can!

katello/katello-change-hostname
johnpmitsch: I'm fine with `capsule list`, since we will run that later.

johnpmitsch (approved):
Works well
```
[root@changed34 ~]# ./katello-change-hostname -u admin -p notmypassword changedagain.example.com

Checking hostname validity

Checking overall health of server

Checking credentials
Invalid username or password
There is a problem with the credentials, please retry
```
and with the actual password
```
[root@changed34 ~]# ./katello-change-hostname -u admin -p changeme$ -yd changedagain.example.com
```
worked as expected :+1:

PR #530, opened 2017-09-14
Fixes #20925 - k-change-hostname should check exit codes

johnpmitsch:
We should check the exit codes on each command ran in the script and exit if they fail.

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -8,8 +8,17 @@ require 'shellwords'
 
 raise 'Must run as root' unless Process.uid == 0
 
+def run_cmd(command, exit_codes=[0], message=nil)
+  result = `#{command}`
+  unless exit_codes.include?($?.exitstatus)
+    STDOUT.puts message if message
+    fail_with_message("Failed '#{command}' with exit code #{$?.exitstatus}")
+  end
+  result
+end
+
 def disable_system_check_option?
-  katello_installer_rpm = `rpm -qa | grep katello-installer`
+  katello_installer_rpm = run_cmd("rpm -qa | grep katello-installer")
   katello_installer_version = katello_installer_rpm[/(\d+\.)(\d+\.)(\d+)/]
   Gem::Version.new(katello_installer_version) >= Gem::Version.new("3.2.0")
 end
@@ -87,14 +96,6 @@ DEFAULT_PROGRAM =  get_default_program
 end
 @opt_parser.parse!
 
-def hammer_ping_success
-  return true if @foreman_proxy_content
-  hammer_ping = `hammer ping`
-  success = $?.success?
-  service_failed = hammer_ping.include? "FAIL"
-  success && !service_failed
-end
-
 def timestamp
   Time.now.strftime("%Y%m%d%H%M")
 end
@@ -166,13 +167,13 @@ def precheck
                       "No changes have been made to your system.");
   end
 
-  STDOUT.puts "\nChecking overall health of server"
-  unless hammer_ping_success
-    fail_with_message("There is a problem with the server, please check 'hammer ping'")
+  unless @foreman_proxy_content
+    STDOUT.puts "\nChecking overall health of server"
+    run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")
   end
 
   STDOUT.puts "\nChecking credentials"
-  hammer_cmd("capsule list")
+  hammer_cmd("capsule list") unless @foreman_proxy_content
   unless $?.success?
     fail_with_message("There is a problem with the credentials, please retry")
   end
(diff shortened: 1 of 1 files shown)


katello/katello-change-hostname
```diff
@@ -256,13 +255,14 @@ unless @foreman_proxy_content
   STDOUT.puts "\nUpdating default #{@proxy}"
   default_capsule_id = hammer_cmd("capsule list | grep #{old_hostname} | awk '{ print $1 }'").chomp
   # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
-  hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null")
+  # For the same reason, we can't use the run_cmd method here
+  hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
```
johnpmitsch: ugg.... this is hacky. This command always errors due to the issue mentioned in the comments above, so we pipe the ouput to `/dev/null/`. I added an exception for the exit code it returns, but this is assuming the exit code is always the same. 

@evgeni @akofink Is there a better way to handle this? And if not, can I rely on the exit code always being the same on this error? This approach worked in my testing.

The tl;dr of [the bug](http://projects.theforeman.org/issues/18186) is that it will update the capsule successfully but hammer returns an error.

evgeni:
` run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")`

this will fail on a proxy. the old code had a unless block around it.

evgeni:
you can add a `|| true` at the end of the command. also hacky. but will always exit 0.

johnpmitsch:
@evgeni thanks, I still have to test on a foreman-proxy!

evgeni:
unrelated, but just realized, is the port really always 9090? shouldn't we read the old port and re use it?

evgeni:
the puppet server/master restart will fail, as it's either puppetserver or httpd (for puppet3) which we have to restart.

johnpmitsch:
> unrelated, but just realized, is the port really always 9090? shouldn't we read the old port and re use it?

@evgeni I'm not sure about the 9090 port, @ehelms or @jlsherrill do you know?

johnpmitsch:
> the puppet server/master restart will fail, as it's either puppetserver or httpd (for puppet3) which we have to restart.

The original cause of adding those lines is [here](http://projects.theforeman.org/issues/18270/). In what scenario is this going to fail?

evgeni:
if you have a Puppet3 setup, the puppet master is running inside Apache httpd using mod_passenger. With Puppet 4 you have the stand-alone puppetserver process. depending on that you have to restart different services. and with run_cmd this will now really fail and wont be silently discarded on a puppet 3 install which has no puppetserver.

ehelms:
I think its 99% safe to assume smart proxies are running on 9090 in Katello deployments.

katello/katello-change-hostname
```diff
@@ -166,13 +167,13 @@ def precheck
...
-  unless hammer_ping_success
-    fail_with_message("There is a problem with the server, please check 'hammer ping'")
+  unless @foreman_proxy_content
+    STDOUT.puts "\nChecking overall health of server"
+    run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")
   end
 
   STDOUT.puts "\nChecking credentials"
-  hammer_cmd("capsule list")
+  hammer_cmd("capsule list") unless @foreman_proxy_content
```
evgeni, line 176: hah, good catch!

katello/katello-change-hostname
evgeni: So we could do
```
hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null || true")
```
here, or keep the 65 exit code as ok.

I think actually keeping 65 is better, as that limits the possibly-accepted errors (and should be stable: https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/exit_codes.rb#L5).

Just for me to understand: it thinks the new URL is invalid, as at the point you are executing this, the service was not yet really moved, right?

evgeni (approved):
ACK if my understanding in https://github.com/Katello/katello-packaging/pull/530/files#r139149900 is correct

johnpmitsch:
@evgeni I'm not sure of the exact cause of updating the capsule w/ hammer working but throwing an error. Its a bad bug and we should fix soon! In my experience, it is consistent. I've tried out of the script too.

PR #532, opened 2017-09-15
Fixes #20944 - kch doesn't check hammer creds on a proxy

johnpmitsch:
Hammer isn't available on a foreman proxy, so we can't use it to check credentials

### DIFF
--- a/katello/katello-change-hostname
+++ b/katello/katello-change-hostname
@@ -170,12 +170,8 @@ def precheck
   unless @foreman_proxy_content
     STDOUT.puts "\nChecking overall health of server"
     run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")
-  end
-
-  STDOUT.puts "\nChecking credentials"
-  hammer_cmd("capsule list") unless @foreman_proxy_content
-  unless $?.success?
-    fail_with_message("There is a problem with the credentials, please retry")
+    STDOUT.puts "\nChecking credentials"
+    hammer_cmd("capsule list", [0], "There is a problem with the credentials, please retry")
   end
 
   if @options[:confirm]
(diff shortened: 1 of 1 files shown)


katello/katello-change-hostname
```diff
@@ -170,12 +170,11 @@ def precheck
   unless @foreman_proxy_content
     STDOUT.puts "\nChecking overall health of server"
     run_cmd("hammer ping", [0], "There is a problem with the server, please check 'hammer ping'")
-  end
-
-  STDOUT.puts "\nChecking credentials"
-  hammer_cmd("capsule list") unless @foreman_proxy_content
-  unless $?.success?
-    fail_with_message("There is a problem with the credentials, please retry")
+    STDOUT.puts "\nChecking credentials"
```
evgeni, line 173: I'd add an empty line before that.

evgeni (approved):
minor nitpick, ACK otherwise!

katello/katello-change-hostname
```diff
@@ -170,12 +170,11 @@ def precheck
...
+    hammer_cmd("capsule list")
```
evgeni: Hah, shouldn't this be

```
run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} capsule list", [0], "There is a problem with the credentials, please retry")
```

as otherwise the unless never happens?

evgeni (changes requested):
eh, found a problem, see above

katello/katello-change-hostname
johnpmitsch: I would rather pass the message to hammer_cmd, and then to run_cmd, so we don't keep repeating the hammer credentials

evgeni:
ack

PR #533, opened 2017-09-18
Fixes #20983 - update method of getting default capsule

johnpmitsch:


katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
   run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
 end
 
+def get_default_proxy_id
+  output = hammer_cmd("--output json capsule list --search=#{@old_hostname}")
```
evgeni: this should be `--search=name=#{@old_hostname}`, right? otherwise it will find `foo-bar.example.com` when you search `bar.example.com`.

katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
   run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
 end
 
+def get_default_proxy_id
+  output = hammer_cmd("--output json capsule list --search=name=#{@old_hostname}")
```
evgeni: even better: `hammer_cmd("--output json capsule info --name '#{@old_hostname}'")`

katello/katello-change-hostname
```diff
@@ -251,10 +258,14 @@ scenario_answers = YAML.load_file("/etc/foreman-installer/scenarios.d/#{@options
...
   STDOUT.puts "\nUpdating default #{@proxy}"
-  default_capsule_id = hammer_cmd("capsule list | grep #{old_hostname} | awk '{ print $1 }'").chomp
-  # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
-  # For the same reason, we accept exit code 65 here.
-  hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
+  proxy_id = get_default_proxy_id
+  if proxy_id
+    # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
+    # For the same reason, we accept exit code 65 here.
+    hammer_cmd("capsule update --id #{proxy_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
```
evgeni: I wonder if we are already touching this, should we also go for replacing the hostname in a saner way?

```ruby
require 'uri'
capsule_uri = capsule['URL']
capsule_uri = URI.parse(capsule_uri)
capsule_uri.host = @new_hostname
capsule['URL'] = capsule_uri.to_s
```

evgeni (changes requested):

katello/katello-change-hostname
johnpmitsch: Wouldn't `capsule` be the old hostname here?

evgeni: `capsule` is the result of `JSON.parse(hammer_cmd("--output json capsule info --name #{@old_hostname}"))`

katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
   run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
 end
 
+def get_default_proxy_id
+  output = hammer_cmd("--output json capsule list --name #{@old_hostname}")
```
evgeni: `info`, not `list`

katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
...
+  proxy = JSON.parse(output)
+  proxy.any? ? proxy.first["Id"] : false
```
evgeni: that won't work anymore. `info` returns a single proxy (if any), so `.any?` and `.first` are not needed.

evgeni: this also means that you'd never return `false` as `hammer_cmd` would error out directly, when the old capsule is not found:

```
[root@sat-r220-02 ~]# hammer --output json capsule info --name sat-r220-02.lab.eng.rdu2.redhat.comee
Error: smart_proxy not found
[root@sat-r220-02 ~]# echo $?
70
```

evgeni (changes requested):

katello/katello-change-hostname
pondrejk: Here `list` should be `info`,  with capsule list hammer says `Error: Unrecognised option '--name'`

pondrejk: nevermind :)

johnpmitsch:
@xprazak2 @evgeni sorry guys, one sec, I have to push to my fork to test, I'll comment when I'm done

johnpmitsch:
Ok, updated

katello/katello-change-hostname
```diff
@@ -245,20 +252,24 @@ if @foreman_proxy_content
...
-  default_capsule_id = hammer_cmd("capsule list | grep #{old_hostname} | awk '{ print $1 }'").chomp
-  # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
-  # For the same reason, we accept exit code 65 here.
-  hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
+  proxy_id = get_default_proxy_id
+  if proxy_id
+    # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
+    # For the same reason, we accept exit code 65 here.
+    hammer_cmd("capsule update --id #{proxy_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
+  else
```
evgeni: This `else` branch will never be called as `get_default_proxy_id` will either return an id or error out itself.

katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
   run_cmd("hammer -u #{@options[:username].shellescape} -p #{@options[:password].shellescape} #{cmd}", exit_codes, message)
 end
 
+def get_default_proxy_id
+  output = hammer_cmd("--output json capsule info --name #{@old_hostname}")
```
evgeni: add `message="Couldn't find default proxy"` to the `hammer_cmd` call to mimick the old `else` path below?

katello/katello-change-hostname
```diff
@@ -237,6 +238,12 @@ def hammer_cmd(cmd, exit_codes=[0], message=nil)
...
+  proxy = JSON.parse(output)
+  proxy["Id"]
```
evgeni: if you return `proxy` instead of `proxy["Id"]` you can use the other details of the proxy (URL) in the calling method :)

katello/katello-change-hostname
```diff
@@ -245,20 +252,24 @@ if @foreman_proxy_content
...
 scenario_answers = YAML.load_file("/etc/foreman-installer/scenarios.d/#{@options[:scenario]}-answers.yaml")
 
 unless @foreman_proxy_content
   STDOUT.puts "\nUpdating default #{@proxy}"
-  default_capsule_id = hammer_cmd("capsule list | grep #{old_hostname} | awk '{ print $1 }'").chomp
-  # Incorrect error message is piped to /dev/null, can be removed when http://projects.theforeman.org/issues/18186 is fixed
-  # For the same reason, we accept exit code 65 here.
-  hammer_cmd("capsule update --id #{default_capsule_id} --url https://#{@new_hostname}:9090 --new-name #{@new_hostname} 2> /dev/null", [0, 65])
+  proxy_id = get_default_proxy_id
+  if proxy_id
```
evgeni: this is always true

evgeni (changes requested):

katello/katello-change-hostname
johnpmitsch: What parts of the proxy url do I need in the capsule update call? Its using the new hostname, and the `get_default_proxy_id` is returning the proxy with the old hostname

evgeni: At the moment, none.
I was still thinking that we can make the update call a bit more flexible, based on the old data. But this does not really belong here into the PR and is an own discussion :)

evgeni (approved):

PR #536, opened 2017-09-20
Fixes #21028 - make k-c-h --help work even without an installed system

evgeni:
/etc/foreman-installer/scenarios.d/last_scenario.yaml might or might not
be there, so let's fall back to `katello` scenario if the file cannot be
found.

### DIFF
--- a/katello/helper.rb
+++ b/katello/helper.rb
@@ -1,7 +1,12 @@
 module KatelloUtilities
   module Helper
     def last_scenario
-      File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
+      last_scenario_yaml = '/etc/foreman-installer/scenarios.d/last_scenario.yaml'
+      if File.exists?(last_scenario_yaml)
+        File.basename(File.readlink(last_scenario_yaml)).split(".")[0]
+      else
+        'katello'
+      end
     end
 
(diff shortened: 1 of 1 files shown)


evgeni:
thanks @theforeman-bot, but f7c3f1c is not the latest commit in this PR, it's 984e955

evgeni:
rebased

katello/helper.rb
```diff
@@ -1,7 +1,7 @@
 module KatelloUtilities
   module Helper
     def last_scenario
-      File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0]
+      File.basename(File.readlink("/etc/foreman-installer/scenarios.d/last_scenario.yaml")).split(".")[0] rescue 'katello'
```
johnpmitsch: is there a way we can avoid a rescue? Maybe check if the file exists?

johnpmitsch (changes requested):

katello/helper.rb
evgeni: Yeah, that might be possible as well.

evgeni:
@johnpmitsch updated!

ehelms:

evgeni:
moved to https://github.com/theforeman/foreman-packaging/pull/1929

PR #537, opened 2017-09-20
Fixes #21029 - define fail_with_message before run_cmd

evgeni:
run_cmd uses fail_with_message, so it has to be defined before


ehelms (approved):

evgeni:
rebased

PR #539, opened 2017-09-22
Update foreman_virt_who_configure to 0.1.7

ares:


ares:
ideally it should be built into 3.4 repos, it's a stable branch build that only fixes bugs of 0.1.x release

ares:
No need to rush, there will be 0.1.7 soon too.

ares:
Updated to 0.1.7, I'll be offline after tomorrow for a while so if there's any change required, please let me know today.

jlsherrill:
ACK

jlsherrill:
I don't think there are any more 3.4 releases, unless something really bad is found, so i will hold off for now

PR #552, opened 2017-10-05
Fixes #21220 - make k-c-h command name easily changed

johnpmitsch:
We need this to backport these changes.

### DIFF
--- a/katello/hostname-change.rb
+++ b/katello/hostname-change.rb
@@ -11,17 +11,18 @@ module KatelloUtilities
   class HostnameChange
     include ::KatelloUtilities::Helper
 
-    def initialize(proxy, plural_proxy, proxy_hyphenated, program=nil, scenario=nil, accepted_scenarios=nil)
+    def initialize(init_options)
       @default_program = self.get_default_program
-      @proxy = proxy
-      @plural_proxy = plural_proxy
-      @proxy_hyphenated = proxy_hyphenated
-      @accepted_scenarios = accepted_scenarios
-
+      @proxy = init_options.fetch(:proxy)
+      @plural_proxy = init_options.fetch(:plural_proxy)
+      @proxy_hyphenated = init_options.fetch(:proxy_hyphenated)
+      @command_prefix = init_options.fetch(:command_prefix)
+      @accepted_scenarios = init_options.fetch(:accepted_scenarios, nil)
       @last_scenario = self.last_scenario
+
       @options = {}
-      @options[:program] = program || @default_program
-      @options[:scenario] = scenario || @last_scenario
+      @options[:program] = init_options.fetch(:program, @default_program)
+      @options[:scenario] = init_options.fetch(:scenario, @last_scenario)
       @foreman_proxy_content = @options[:scenario] == @proxy_hyphenated
 
       setup_opt_parser
@@ -77,7 +78,7 @@ module KatelloUtilities
     def get_fpc_answers
       register_in_foreman = false
       certs_tar = @options[:certs_tar]
-      " --foreman-proxy-register-in-foreman #{register_in_foreman} --#{@proxy_hyphenated}-certs-tar #{certs_tar}"
+      " --foreman-proxy-register-in-foreman #{register_in_foreman} --foreman-proxy-content-certs-tar #{certs_tar}"
     end
 
     def precheck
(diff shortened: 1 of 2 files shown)


katello/katello-change-hostname
```diff
@@ -9,7 +9,14 @@ end
 
 require File.join(UTILS_PATH, 'hostname-change.rb')
 
-hostname_change = KatelloUtilities::HostnameChange.new("Foreman Proxy", "Foreman Proxies", "foreman-proxy-content")
+options = {
+ proxy: "Foreman Proxy",
+ plural_proxy: "Foreman Proxies",
+ proxy_hyphenated: "foreman-proxy-content",
+ plugin: "katello"
```
ehelms: I think I might prefer this being `command` or `command_name` rather than plugin. Plugin tends to evoke a lot of other thoughts and this isn't really a plugin system.

johnpmitsch:
@ehelms updated to `command_prefix`, I think that makes sense for what we are trying to achieve

katello/hostname-change.rb
```diff
@@ -11,17 +11,18 @@ module KatelloUtilities
...
+      @proxy_hyphenated = init_options[:proxy_hyphenated]
+      @accepted_scenarios = init_options[:accepted_scenarios]
+      @command_prefix = init_options[:command_prefix]
       @last_scenario = self.last_scenario
+
       @options = {}
-      @options[:program] = program || @default_program
-      @options[:scenario] = scenario || @last_scenario
+      @options[:program] = init_options[:program] || @default_program
+      @options[:scenario] = init_options[:scenario] || @last_scenario
```
ehelms: Hashes have a builtin function for this `.fetch(:scenario, @last_scenario)`

katello/hostname-change.rb
```diff
@@ -11,17 +11,18 @@ module KatelloUtilities
...
-      @proxy = proxy
-      @plural_proxy = plural_proxy
-      @proxy_hyphenated = proxy_hyphenated
-      @accepted_scenarios = accepted_scenarios
-
+      @proxy = init_options[:proxy]
+      @plural_proxy = init_options[:plural_proxy]
+      @proxy_hyphenated = init_options[:proxy_hyphenated]
+      @accepted_scenarios = init_options[:accepted_scenarios]
+      @command_prefix = init_options[:command_prefix]
```
ehelms: If these are all required, you can use `fetch` to enforce that.

johnpmitsch:
@ehelms thanks, updated

katello/hostname-change.rb
```diff
@@ -11,17 +11,20 @@ module KatelloUtilities
...
+      # accepted_scenarios is not a required parameter, so we allow to return nil.
+      # See method in helper file
+      @accepted_scenarios = init_options[:accepted_scenarios]
       @last_scenario = self.last_scenario
+
       @options = {}
-      @options[:program] = program || @default_program
-      @options[:scenario] = scenario || @last_scenario
+      @options[:program] = init_options[:program].fetch(:program, @default_program)
+      @options[:scenario] = init_options[:scenario].fetch(:scenario, @last_scenario)
```
ehelms: I think you mean to leave off the [:...] for these two options?

johnpmitsch: updated

katello/hostname-change.rb
```diff
@@ -11,17 +11,20 @@ module KatelloUtilities
...
-      @proxy_hyphenated = proxy_hyphenated
-      @accepted_scenarios = accepted_scenarios
-
+      @proxy = init_options.fetch(:proxy)
+      @plural_proxy = init_options.fetch(:plural_proxy)
+      @proxy_hyphenated = init_options.fetch(:proxy_hyphenated)
+      @command_prefix = init_options.fetch(:command_prefix)
+      # accepted_scenarios is not a required parameter, so we allow to return nil.
+      # See method in helper file
+      @accepted_scenarios = init_options[:accepted_scenarios]
```
ehelms: You can make this explicit without the need for comment by using `init_options.fetch(:accepted_scenarios, nil)` which will set the default to nil.

johnpmitsch: thanks, updated

ehelms (approved):

PR #553, opened 2017-10-05
Fixes #21221 - disable system checks in katello-* scripts

johnpmitsch:


ehelms (approved):

evgeni:
I agree in the hostname-change case, but a restore can happen on a different system, and that can potentially fail the requirements, no?

johnpmitsch:
@evgeni my thought is that the installer would already have been run since katello is already installed for both script cases. If katello has already been installed, the user has either decided to skip the system checks or has passed them. I don't see a reason to check them again.

evgeni:
Right, because we are restoring to an empty, but installed katello... ACK!

evgeni (approved):

PR #556, opened 2017-10-11
fixes #21270 - custom certificates added to backup

cfouant:


katello/backup.rb
```diff
@@ -67,13 +68,19 @@ def configure_configs
...
         "/etc/candlepin",
         "/etc/foreman",
         "/etc/hammer",
         "/etc/sysconfig/tomcat*",
         "/etc/tomcat*",
         "/var/lib/candlepin",
+	scenario_answers["certs"]["server_cert"],
+	scenario_answers["certs"]["server_key"],
+	scenario_answers["certs"]["server_cert_req"],
+	scenario_answers["certs"]["server_ca_cert"]
```
sean797: is it just me or is the alignment off ?

cfouant: :+1: great catch, fixing now

cfouant: Actually @sean797 it's correct, just looks strange here. If you look at "raw" code without the markup, you'll see it is aligned correctly.

sean797: Its because you've used a tab 😈 please use spaces

johnpmitsch:
Works on a local setup w/o custom certs for both server + capsule, but I would like to test this on a system with custom certs. The only one I have access to isn't available today, but I can use it next week, I'll test and comment back here

johnpmitsch (approved):

PR #558, opened 2017-10-16
FIxes #21341 - check for /usr/sbin path of the tools when loading

evgeni:


johnpmitsch:
@evgeni I didn't realize you had already opened this (much appreciated) when I opened up this https://github.com/Katello/katello-packaging/pull/559

johnpmitsch:
@evgeni why not just check if the file path includes `sbin`? That seems  unique enough to test if its production or not. The only edge case is having some tool with a substring of "sbin", which seems unlikely and even if it happens, we are still defaulting to production working (and dev setup breaking)

evgeni:
@johnpmitsch yeah, either works. closing as yours is merged.

PR #580, opened 2018-02-06
Bump rubygem-katello to 3.5.1.1

akofink:


akofink:
@ehelms This one should be good to go.

PR #582, opened 2018-04-02
3.5.2 Cherry Picks from foreman-packaging

akofink:
@jturel @johnpmitsch Care to review your commits? This didn't really cherry pick cleanly, probably because it's from a completely different repo.

### DIFF
--- a/katello/backup.rb
+++ b/katello/backup.rb
@@ -128,10 +128,8 @@ module KatelloUtilities
       @dir = File.join directory, "#{@program}-backup-" + self.timestamp unless @options[:no_subdir]
       puts "Creating backup folder #{@dir}"
       FileUtils.mkdir_p @dir
-      unless @options[:no_subdir]
-        FileUtils.chown_R nil, 'postgres', @dir if @databases.include? "pgsql"
-        FileUtils.chmod_R 0770, @dir
-      end
+      FileUtils.chown_R nil, 'postgres', @dir if @databases.include? "pgsql"
+      FileUtils.chmod_R 0770, @dir
     end
 
     def snapshot_backup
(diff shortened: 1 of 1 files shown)


johnpmitsch:
@akofink https://github.com/Katello/katello-packaging/pull/582/commits/3bf96ae9490eb3f6eff3af0737884bbf7ee9cc40 looks fine to me

katello/helper.rb
```diff
@@ -28,28 +40,63 @@ def disable_system_check_option?
...
 
     def fail_with_message(message, opt_parser=nil)
-      STDOUT.puts message
+      STDOUT.puts message.red
       puts opt_parser if opt_parser
       exit(false)
     end
 
-    def run_cmd(command, exit_codes=[0], message=nil)
+    def fail_if_file_not_found(files)
```
jturel: How close are we trying to get with the cherry pick? Asking because it looks good for the most part but this `fail_if_file_not_found` does not look familiar to me, so it probably belongs to another PR and not the one you've cherry picked

johnpmitsch: That is from this commit here https://github.com/theforeman/foreman-packaging/commit/ea534f6227781721bfcc9922b72c0d77ee1e9b69

@akofink are the custom cert options for k-c-h included in 3.5?

johnpmitsch:
Once we get the cherry-picks straightened out, it would be a good idea to run these commands manually. As far as I know, we don't do this in  the release pipeline (another good part about moving  these to foreman-maintain!)

katello/helper.rb
akofink: Weren't they added in https://github.com/theforeman/foreman-packaging/commit/ea534f6227781721bfcc9922b72c0d77ee1e9b69, or are there other code changes required?

akofink:
Kicked http://projects.theforeman.org/issues/21517 from 3.5.2, so now it's just the one issue.

johnpmitsch:
@akofink looks like its missing [the first change here](https://github.com/Katello/katello-packaging/commit/3bf96ae9490eb3f6eff3af0737884bbf7ee9cc40#diff-e5c9a75a6cce52c183d1d26a2500311eL131)?

akofink:
Rogue cherry-pick! Good catch. Updated. :)

johnpmitsch (approved):

